PR #200, opened 2020-07-28
Dia

oivannikov:
[DEMO LINK](https://oivannikov.github.io/layout_dia/)


irenhh:
scroll-behavior

=====

all interactive elements should have cursor pointer, you must do it automatically without a reminder

=====

image
those are buttons, add corresponding styles

=====

image
this is a button as well

=====

image
why is it black?

=====

form should be submitted with 'enter' key, make sure your buttons is of type 'submit'

=====

image
pointer

=====

image
do you have something else purple in your design?

=====

image
tablet: why there is so much space under the text, move it to the center

=====

image
fix indents

=====

image
don't make it transparent

=====

image
too much empty space

=====

image
would be nice so that button has the same width as the form

=====

image
add equal indents between icons

=====

image
mobile: fix according to the design

=====

image
add some spaces on the edges

=====

image
what happened with ratio?

=====

please pay more attention to the quality of your work

irenhh:
image
add hover to this buttons as well, it is the same as others

=====

image
why the font color is different depending on focus state, it should be stable

=====

image
why is button is outside the form?

=====

src/styles/section.scss
```diff
@@ -24,6 +24,7 @@
     margin-top: -90px;
     border-radius: 30px 0 0 30px;
     padding: 204px;
+    scroll-behavior: smoothpx;
```
irenhh, line 27: ?

irenhh: scroll-behavior is usually added to html tag

irenhh:
did you deploy changes?

irenhh:
approved, but you need to fix horizontal scroll

PR #281, opened 2021-09-28
First Variant

misharosa:
-DEMO[https://misharosa.github.io/layout_dia/]


VinnieJ-2k20 (changes requested):
Good start, but your work needs some fixes.

1. Seems like you forgot to set max-width for the content, on 3000px the website is not looking good.
image

2. Check the font style here with the design
image

3. Background of "Who we are" section should be different
image

4. Same here
image

5. Distance between these elements in footer is different from the design
image

6. Need to fix form styles, incuding autocomplete
https://mateacademy.slack.com/archives/C02CSGP1KD1/p1633343754071500
image

7. On mobile version distance here is different
image

8. Make hover effect on these more simple
image

VinnieJ-2k20 (changes requested):
Okay, that's better.
But some of the mentioned things were not fixed:
1. Here you can still see white background in the corner - better to set it default to the body.
image

2. On large screens content is still too wide
image

3. On ~1650px "What people say" section is not centered and is causing horizontal scroll for the page
image

4. Form still has default styles with autocomplete
Fix: https://mateacademy.slack.com/archives/C02CSGP1KD1/p1633343754071500
image

5. Distance from form to nav menu in footer was not fixed.
image

6. I see lots of commented code. It should not be present in your project.

VinnieJ-2k20 (changes requested):
Okay, it's better, but not ready yet.

1. You fixed it on mobile, but now it's broken on desktop
image

2. Font here is still different from the design. Check the font-family
image

3. Check again width limit for the content. It SHOULD NOT be bigger than on the design
image
image


4. Form styles should look like the design even with autocomplete. This is incorrect
image

VinnieJ-2k20 (changes requested):
Check Slack for details.

VinnieJ-2k20 (changes requested):
1. Text here is centered differently. Also, check the size and border-radius of the button.
image

2. With opened menu, scroll is still enabled.
image

3. Add some padding here for small screens
image

4. Here there should be bigger distance between sections on mobile screens.
image
image

5. Here things are centered differently than on the design
image

PR #285, opened 2021-10-06
solution

ivarovskiy:
[DEMO](http://ivarovskiy.github.io/layout_dia/)


VinnieJ-2k20 (changes requested):
Layout loks good.

1. Add hover-effects (you already know that)
2. Fix form styles on autocomplete and placeholder in textarea. How to do it - look at pinned messages in slack group chat.
image

3. Fix header on wide screens
image

4. Not the best layout for burger menu. Make it extend vertically
image

PR #301, opened 2021-11-01
part of mobile header only

i-stanko:


src/index.html
```diff
@@ -1,13 +1,150 @@
...
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link
+      rel="shortcut icon"
+      href="images/logo/air-logo.svg"
+      type="image/x-icon"
+    >
     <title>Dia</title>
```
IrynaHerasymchuk, line 11: Change it to 'Air'

src/styles/blocks/about-us.scss
```diff
@@ -0,0 +1,43 @@
+.about {
+  padding: 120px 0;
+  //background-color: yellow;
```
IrynaHerasymchuk, line 3: Don't forget to remove this later :)

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,24 @@
...
+  display: block;
+  width: 280px;
+  height: 50px;
+  border-radius: 8px;
+  background-color: $colorBlue;
+  font-weight: 600;
+  font-size: 15px;
+  line-height: 50px;
+  text-align: center;
+  transition: all $transitionDuration;
```
IrynaHerasymchuk, line 11: Try to avoid using 'all' for transition, because it has bad effect on browser, it's better to state every property that you want to have transition on

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+    align-items: center;
+    justify-content: space-evenly;
+    width: 100%;
+  }
+
+  &__item {
+    position: relative;
+    display: flex;
+    justify-content: center;
+    height: 100%;
```
IrynaHerasymchuk, line 24: Would be better if you add some space between link and after element

src/styles/blocks/top-actions.scss
```diff
@@ -0,0 +1,26 @@
...
+    align-self: center;
+
+    @include onDesktop {
+      opacity: 0;
+      cursor: none;
+    }
+  }
+
+  &__menu-closer {
+    background-image: url("../images/cross.svg");
```
IrynaHerasymchuk, line 20: Your icon is repeating itself now, remove it

IrynaHerasymchuk (approved):
Not bad!

ihor-jpeg (changes requested):
Good work, but there are a few things to work on.

- when I open the menu and scroll, i can still see the images and other elements of the website that i should not see
Screenshot 2021-11-03 at 11 52 01

- the Air logo seems to be offset
Screenshot 2021-11-03 at 11 56 37

- text here seems too narrow, i think you have misspelled the name of your grid item class for that particular element
Screenshot 2021-11-03 at 11 54 51

- the card seems to be narrower that it should be
Screenshot 2021-11-03 at 11 57 38

- when I press the 'Send' button nothing happens
Screenshot 2021-11-03 at 11 45 56

ihor-jpeg (changes requested):
There are still some stuff to work on

-you need to add buttons to this section
Screenshot 2021-11-04 at 11 38 20

- the cards are way to skinny here
Screenshot 2021-11-04 at 11 36 58

other than that, good work!

aionova (approved):
GJ!

PR #349, opened 2022-01-12
dia solution

oleksandr-kozhushko:
- [DEMO LINK](https://oleksandr-kozhushko.github.io/layout_dia/)

### DIFF
--- a/.bemlintrc.json
+++ b/.bemlintrc.json
@@ -1,0 +1,16 @@
+{
+  "elementDivider": "__",
+  "modifierDivider": "--",
+  "ignore": [
+    "node_modules",
+    "dist"
+  ],
+  "rules": {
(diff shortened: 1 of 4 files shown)


Anna-Tkachenko (changes requested):
Good job!
Check some points:
1) It's better not to reload page when form submitted as we did in miami langing
2) Check 'services' title design in figma
image
3) Check your mobile version, it has some interesting design
image

Anna-Tkachenko (approved):
Well Done!
Some comments:
1) Add styles on hover state for links in footer (call us link, visit us link)
2) You didn't fix this text positioning according to design
image
3) Check form width with design, on tablet screens it aligned by left and doesn't have 100% width

PR #351, opened 2022-01-19
solution

013g:


Anna-Tkachenko (changes requested):
Good job!
Check some points:
1) Add favicon to your page
2) It's better to add scrolling to some block by clicking on 'learn more' button in header
3) Check max-width for this text
image
4) Check all indents in your form, I think it's bigger than in design
image
5) Visit us link in footer should be opened in the new tab
6) Same with social links in footer, it should be opened in the new tab
7) Check max-width of this block, it's the same in main cases
image
8) It's better not to reload page on form submitting as you did in miami landing

Anna-Tkachenko (approved):
Some comments:
1) Check this lines near social links in footer, delete them
image
2) Check where you added <script> tag in your file, it should be before closing body tag for avoiding such errors
image
3) It's better not to scroll to the top of the page when form is submitted

PR #354, opened 2022-01-28
write code

Batick3737:


polosanya (changes requested):
Keep up the good work! 
1. Let's add a favicon for the page
image
2. Add hover effects for all interactive elements
image
3. Images don't match the design. Also, add increasing hover effects for all images
image
image
4. After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
image
5. Form shouldn't reload the page and submit data through URL
image
6. Add hover effects.
Social links should be links to some external pages.
Also, make sure that all external links are opened in a new tab (google maps, social medias etc)

Batick3737:
> 7. According to Figma social media icons should be blue on hover
> 
> Screenshot 2023-05-30 at 21 52 06

В мене немає, активних картинок і не знаю, як зробити їх синім

witflash (changes requested):
Let's improve a little bit.

1. You can use transparent background and on hover change the color. It should prevent "jumping" button text
image
image

2. Button "Apply" should lead somewhere. Add some link or an anchor
image

3. It's better to use blue color for socials hover, I think
image

4. Could you use JS code and disable page scroll when your menu is opened?
image

loralevitska (changes requested):

- fix this text, it should be on one line
image
- add smooth transition for all elements with hover effect
- fix `405 Not Allowed` error after form submit. Form should just send data, and clear fields after that
- open this link on new tab 
image
- fix footer text according design on mobile mode
image
- center these blocks according mobile mode design 
image
- change logo name to `Air` according design
image

PR #381, opened 2022-03-23
Develop

ElizaChernysh:
[Link](https://elizachernysh.github.io/layout_dia/)


MaksimusCRS (changes requested):
Не работает адаптив для секций What people say и футера

Anna-Tkachenko (changes requested):
Good job!
Some comments:
1) Don't forget to add favicon to your page
2) Check design of this part of header in figma
image
3) Bottom blue lines should be the same for all menu links
image
4) Check design of this section (block positioning) in figma
image
5) All fields in the form should be required
6) Check design and positioning of this section in figma
image
7) I can scroll the page when menu is opened, fix it
image

PR #390, opened 2022-04-12
Dia Start

WojciechCzarnowski:
[DEMO LINK](https://WojciechCzarnowski.github.io/layout_dia/).


ihor-jpeg (changes requested):
- let's add transition on buttons, social media icons and header links
- i can submit an empty form
- social media icons are not of the same color
Screenshot 2022-04-21 at 13 05 03

ihor-jpeg (approved):

PR #392, opened 2022-04-18
not a solution

yelnikov-andrii:
https://yelnikov-andrii.github.io/layout_dia/


Anna-Tkachenko (changes requested):
Good job! 👍 
Some comments:
1) Check correct sizes for header in figma (check bottom indents for main title)
2) Check correct position for this block
image
3) Check correct indents in this block (the height of this block should be bigger according to design)
image
4) Check the design for the form
5) Form shouldn't be submitted empty
6) On form submitting I see 405 error - fix it
7) So big indents in footer - fix it
image
8) Social icons in footer should be the links on social networks (don't forget to open them in the new tab)
9) Same with address link in footer - you should open it in the new tab
10) I can't open mobile menu - fix it

Anna-Tkachenko (changes requested):
Some comments:
1) check form design in figma
image
2) Try to center all icons
image
3) Address link in footer should be opened in the new tab
4) Check correct design for menu in footer for small screens
image
5) If I scroll the page with opened menu I see elements from other sections - fix it

Anna-Tkachenko (approved):
Well Done! ✔️ 
Just try to refactor your footer, check with design correct width and positioning of this two blocks
image

Anna-Tkachenko (commented):
Also will be better to add 'Your email' as placeholder for this input
image

PR #399, opened 2022-04-25
dia

ISRLv2021:
[DEMO LINK](https://ISRLv2021.github.io/layout_dia/).


Anna-Tkachenko (changes requested):
Good job! 👍 
Some comments:
1) Social icons in footer should be the links on social networks (don't forget to open them in the new tab)
2) Address link should be opened in the new tab
3) Check correct font styles here
image
4) Check correct indents between form fields
5) Check correct width for this blocks
image
6) Check correct font styles here
image
7) Check correct font styles
image
8) Fix opened menu
image
9) Will be better to set some max-width for this blocks
image

PR #410, opened 2022-05-27
Added first three blocks (mobile only)

DenysPolitov:


jstmpelowycz (changes requested):
https://www.loom.com/share/0d9eb6c056ee4e4b95d52b074b2dcafb

PR #414, opened 2022-05-30
few blocks of Dia were added

dima-pyvovarchuk:
- [DEMO LINK](https://dima-pyvovarchuk.github.io/layout_dia/)


mykola-koval (approved):
- maybe add some hover effects on these photos
image

 - textarea in form should also be required

PR #415, opened 2022-05-31
header

Ihor-Novakovskyi:
https://ihor-novakovskyi.github.io/layout_dia/


illiarerun (approved):
GJ, you should disable resize for textarea
image

PR #430, opened 2022-06-30
solution

tumann21:
[DEMO LINK](https://tumann21.github.io/layout_dia/)


aliceblunt8 (changes requested):
- add hover effects to these links
Screenshot 2022-07-01 at 15 59 04
- All the social icons in the footer should be clickable and open the social networks in a new tab
- Logos in header and footer should be links to home page
- Make sure when you click on logo there is no 404 error (Use href="#" in anchor tag)

Bohdan-Maksymiuk (approved):

PR #452, opened 2022-07-28
solution

AndrejGutnik:
https://AndrejGutnik.github.io/layout_dia/


ArturErmolenko (changes requested):
1. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
2.Add some hover effects and make it links to social(Facebook insta...)
Screenshot 2022-08-02 at 11 07 40

marina-tilniak (changes requested):
варто змінити дефолтні стилі при автокомпліті [(читай тут)](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)

Virass (approved):
Good job! 👍

PR #457, opened 2022-07-29
Develop

serhii-yunnikov:
[DEMO LINK](https://serhii-yunnikov.github.io/layout_dia/).


marina-tilniak (changes requested):
- відправка форми не повинна перезавантажувати сторінку(https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
- потрібно додати hover ефекти на зображення та посилання
- зовнішні посилання (на google maps наприклад) мають відкриватися у новій вкладці
- варто додати якийсь Х для виходу з меню
image

Bohdan-Maksymiuk (changes requested):
Good job 👍 
Fix some comments ⬇️ 

Screenshot 2022-08-15 at 22 41 03

* check this section on design

Screenshot 2022-08-15 at 22 43 47

* check background color here

valerii-derkach (approved):

polosanya (changes requested):
1. I think it would be great for main screen to take the whole height, so on large screens it looks okay too
image
2. Don't forget to add hover effects for all interactive elements
image
3. Make this text clickable too. Add cursor pointer on hover for links
image
4. It would be great to add some validation for the form, so user can't submit empty form or invalid email. Also, cursor pointer is absent for submit button
5. Add cursor pointer for these icons
image
6. Add more distance between footer links on mobiles
image
image

agniya-i (changes requested):
1. Fix slider on tablet, also check border-radius, it should be for all angles 
Screenshot 2023-03-13 at 19 03 40

2. Fix form, it shouldn't open the http://localhost:8080 and reload page

anstsot (changes requested):
Great job, but there are few fixes:
1. after submitting a form the page is reloading. it should just reset the form and stay on the same place.
image

2. the titles are not visible in footer
image

3. it would be great to add bigger margin here
image
image

qoa11a (changes requested):
image
The titles still aren't visible in the footer.

PR #468, opened 2022-08-19
Completed header without sidebar

Proph7000:
[DEMO LINK](https://proph7000.github.io/layout_dia/)


liliia-livandovskaya (changes requested):
GJ 👍 🔥  Please fix your code according to the next points:
1. add a favicon 
2. Apply :hover effect for images on the page and on your menu. The more effects your page would have more chances that HR would like your portfolio.
3. Fields from the form should be reseted after submitting
4. On the screen 2560px your header is looking weird 
5. When opening the menu it shows too fast. The speed of animations should be the same throughout the landing page

liliia-livandovskaya (approved):
Great job! 👍  Let's add the hover effect to the TESTIMONIALS block too.

PR #478, opened 2022-09-03
ver1.0

PaulVoron:
[DEMO LINK](https://PaulVoron.github.io/layout_dia/)


liliia-livandovskaya (changes requested):
Great job! please pay attention to the following comments:
1. Transition duration should be the same for every change.  For example, when I hover over icons on the footer, the color change is too fast. Fix it all over your page
2. After submitting the form page should not be reloaded
3. Add the link to google maps in the footer 
Знімок екрана 2022-09-05 о 12 01 49

liliia-livandovskaya (approved):
Great job 👍  your landing is awesome and that is a big plus that you have added a slider to your page

PR #483, opened 2022-09-12
Dev

fatihsamur:
[DEMO LINK](https://fatihsamur.github.io/layout_dia/)


kamal-tufekcic (approved):
Very functional and responsive, only suggestion I have is to add some animations to certain elements, apart from the buttons, the whole site feels very static

PR #486, opened 2022-09-14
v01

Artem193:
https://Artem193.github.io/layout_dia/


SanyaBratashchuk (changes requested):
1. Make scroll smooth
2. Hover effects shouldn't affect section height
3. Transitions should work in a both ways 
4. Should be link to real socials
image
5. Form shouldn't refresh the page 
6. Make fields required

SanyaBratashchuk (changes requested):
1.Make this buttons using styles, to avoid loses of quality also you will be able to add some hover effects to them
image
2. Add validation to the email field 
3. use svg
image
4. Add hover effects to the links in contacts and navigation sections

PR #491, opened 2022-09-21
Develop

Sergey-Mironenko:


Sergey-Mironenko:
https://Sergey-Mironenko.github.io/layout_dia/

SanyaBratashchuk (changes requested):
1.hover effect is too big
image
2.link is not working 
image
3.Should lead to the real socials 
image
4.Add transitions

AlisaVlasova (changes requested):
social media icons should become blue on hover, now its only twitter by default
image

fix this behavior on slide change and add some hover effect on arrow buttons
https://user-images.githubusercontent.com/66749919/193399517-b4537c12-6043-4a44-afd1-014445b3f0a8.mov

https://user-images.githubusercontent.com/66749919/193399613-53975424-f913-469a-83a2-1d61fd2f0a15.mov

mobile menu doesn`t work
remove horisontal scroll on mobile
image

PR #496, opened 2022-10-03
Develop

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/layout_dia/)


liliia-livandovskaya (changes requested):
Great job! There are several things to work on:

1. Remove this 
Screenshot 2022-10-05 at 15 48 36

2. Page should not be reloaded after the form submit 

3.  Menu does not open
Screenshot 2022-10-05 at 15 49 09

qoa11a (changes requested):
Great job! Few comments:

image
image

- When I focus on the message textarea it becomes bigger. It shouldn't unless it needs to fit content inside of it.
- Form shouldn't scroll user to the top of the page after submission.
- Please either provide correct example of email or use something like "Your email" as a placeholder. What you use as placeholder now is value of the input on the design, not a placeholder.

IrynaK-a:
> * orm shouldn't scroll user to the top of the page after submission.

Sorry, but If I understand clearly requirement 9 want me to scroll it:
9. When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded.
Should it be?

And could you give me an advice in order to requirement 20:
20. The user must have the opportunity to conveniently write 2-3 lines of text in the message field.
How can I do it only if I need to fit content?

Thanks.

PR #501, opened 2022-10-17
Air was created

katya-suvorova:
https://katya-suvorova.github.io/layout_dia/


loralevitska (changes requested):
good job! need improvements

1. remove background-color 
image
2. your page shouldn't break after click on logo. Check it on all logos. 
image
I clicked here 
image
3. Change title style, because now with the logo it looks like the title was duplicated
image

4. add some indentation for harmony. I think 30-40 pixels will be enough
image

tetiana-ivanchenko (changes requested):
1. your form shouldn't reload the page, also it shouldn't show any data in the link
image
2. transition should work not only when I move the mouse over the button, but also when I remove it (it should work for all hover effects)
image
3. on tablet this block don't match to design
image

liliia-livandovskaya (changes requested):
Great job! Your landing looks awesome 🔥 

But there should not be an error after submitting the form. Please fix it 

Screenshot 2022-12-06 at 15 51 04

polosanya (changes requested):
Almost perfect!
1. Would be great to add some hover effects to these links
image
2. This button has the wrong font (Arial instead of Poppins). Please fix it according to the design
image
3. IMHO, it is better to use the same font in the burger menu as in the whole projectimage

loralevitska (approved):

PR #507, opened 2022-10-28
air

panianchuk:
https://panianchuk.github.io/layout_dia/


SanyaBratashchuk (changes requested):
Check font-family
image
Add spacing between this buttons 
image
email field shouldn't be initially red
image
Page shouldn't refresh after form submit

SanyaBratashchuk (changes requested):
make fields required

DarkMistyRoom (changes requested):
- It looks not good
image


- It is a result of clicking on the link. Unnecessary space on the top
image

- Slider on the header section doesn't work

Bohdan-Maksymiuk (approved):
GJ

OleziO (changes requested):
Demo:
image
Design: 
image


Make animation slower 
image

Design: 
image
Demo: 
image

PR #510, opened 2022-11-01
Landing Air was created

serge-kontakt:
[DEMO LINK](https://serge-kontakt.github.io/layout_dia/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -365,7 +365,7 @@
 
             <div class="footer__title" id="contact-us">Send us  a message</div>
 
-            <form action="https://serge-kontakt.github.io/layout_dia/" method="post">
+            <form action="#">
               <input
                 name="email"
                 type="email"


polosanya (changes requested):
Great work! Just found some moments it would be great to fix
1. Use the POST method for your form. so that the data is not transmitted via url
image
2. Make your hover effects a bit smoother. Use transition
3. Let's make these links navigate to some social media pages
image
image
4. Button is not centered on mobile version
image
4. Same here
image

serge-kontakt:
@polosanya  Thank you!
I've fixed almost everything.
Regarding #1"POST" - there was not such task. If I put POST then I have an ERROR because of Action. Or give me an advice what I should to put in Action with POST.

qoa11a (approved):
You can add method="post" and action="", then call event.preventDefault()

PR #511, opened 2022-11-02
DIA FINISH (Fixed problems)

Igor-Yaremchuk:
[DEMO LINK](https://Igor-Yaremchuk.github.io/layout_dia/)


qoa11a (changes requested):
- You can make inputs in form required and specify type="email" for basic email validation.
- Check the background-color of the body in design and compare it to yours. They don't match.

PR #518, opened 2022-11-14
Air

Michael1195:
[DEMO LINK](https://Michael1195.github.io/layout_dia/)


qoa11a (changes requested):
image
- Let's make header and footer be full width on big screens. I mean their background-color.
- Let's disable page scroll when mobile menu is open. And open it in full screen.

polosanya (changes requested):
Looks really great! Just found some small moments that would be great to fix
1. Please check the 13th point of the checklist
image
image
2. Add some [active](https://developer.mozilla.org/en-US/docs/Web/CSS/:active) effect to the buttons, not only hover
button

agniya-i (approved):
👍

PR #521, opened 2022-11-15
Develop

NataliiaHen:
https://NataliiaHen.github.io/layout_dia/


SanyaBratashchuk (changes requested):
1. Fix transitions 
image
2. Textarea has a different font check it 
image
3. This space looks strange 
image
4. Form shouldn't lead to the top fo page after submit
5. Do you have menu for mobile?
image

TarasHoliuk:
1. The form and buttons should be full-width on mobile screen (check mockup):
image

2. Hire Us is not aligned
image

3. Make it the same for all sections (check the mockup):
- Testimomials - header is full-width, cars are not:
image

- Services - both with horizontal padding:
image

And checklist points 11 and 14

TarasHoliuk (changes requested):
You did an amazing job :fire: 

But I've added a few small comments to make it perfect (mostly about alignment)

loralevitska (approved):
Well done) Check my comments for improvement 
- make sure that all interactive elements have a cursor pointer on hover (nav links on page and in menu, learn more buttons etc)
- also would be better if cross icon in menu will be at the same position with burger menu icon

PR #526, opened 2022-11-22
Layout-Air

92BS:
[DEMO LINK](https://92BS.github.io/dia-landing/)
[Link to Figma Design] (https://www.figma.com/file/7qwsWggv9BAxMi2VPhBuPr/Air-(formerly-Dia))?node-id=22904%3A672&t=QQ0tAHGyeGsLU270-0)


Anastasiia-Svintsova (changes requested):
Your demo link is not working
Animation

liliia-livandovskaya (changes requested):
Great job! There are a couple of things to work on:

1. The link should be opened in the new tab, not in the same
Screenshot 2022-12-05 at 18 17 03

2. Add a hover effect on this block 
Screenshot 2022-12-05 at 18 17 18

loralevitska (changes requested):
almost done) 
1. remove this background 
image
2. hover effect to links not fixed 
image
3. this block should navigate to to the corresponding block. I see in the url `#contact-us` but nothing happens when I click. I guess it should be `#contacts`
image

polosanya (changes requested):
The page looks like this and scroll is absent, so I can't check it properly.
The main screen should take all the height.
Please fix it
image

Anastasiia-Svintsova (changes requested):
- Before clicking on any section, the page scroll is not available. Fix it please

https://user-images.githubusercontent.com/99403198/209540166-b9a36451-8f87-4e87-89ac-1a337af0d0cd.mov

- Seems that you have different fonts in the form. Fix it please. 
image

- This links should be opened in a new tab
image

92BS:
Add changes :
-target blank to social icons to be opened in new tab
- Add font-family to "Send us message" form
- Fix scroll, but need to check from mentor's side

polosanya (changes requested):
1. The main problem is still not fixed(
image
2. Would be great to add some hover effects for the whole cards, not just arrows
image
3. Let's add some hover effects to the buttons. Also, seems like the font doesn't match the design
image
4. After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/).
Btw, the text styles in placeholders are still different
image
5. Add hover effects to the links and open social medias in a new tab
image
6. Font doesn't match the design
image
image
7. Let's center the links horizontally. The cross icon should be in the right corner 
image

loralevitska (changes requested):
1. Almost done)
image
You're left with a problem hovering over these 4 elements.
Transform properties are overwritten if they are not written on the same line. So you need to rethink your decision
You have 2 ways: either do not use the transform property to move the blocks, or use other hover effects (I would choose another option, you can use a background-shadow)

qoa11a (approved):

PR #527, opened 2022-11-24
created dia

KseniyaGonchar:
https://KseniyaGonchar.github.io/layout_dia/


DarkMistyRoom (changes requested):
- Logo shouldn't link to 404 error

- The header should fit the window height
https://www.loom.com/share/783b32c3494f4729b81c6a04090e0619

- Add a hover effect to all images, links, and buttons

- Form fields shouldn't change the background on typing
image

- The cursor should became a pointer on hovering over this button 
image

- Page shouldn't reload on the form submitting
- These links should lead to real social networks
image

- Header should contain slider with these photo
image

- Fix it
https://www.loom.com/share/d58577800e4641748f2cbccbf9c20bc3

- Remove this line from the smaller screens
image

KseniyaGonchar:
thanks, have tried to fix, but i have troubles with slider

KseniyaGonchar:
i don`t know why but picture on my screen is different than a picture when i open link here
Снимок экрана 2022-12-01 в 08 29 11

vlad-demchuk (changes requested):
1. add a favicon
2. add hovers
image
3. use svg for icons fot better quality
image
4. text is cut
image

KseniyaGonchar:
I can't figure out why it's cut out, on my screen I can see it in full 
thanks, will try to understand

AlisaVlasova (changes requested):
Good job! Almost there)
1. on a large screen after 1440 media queries for this block don't work
image
2. cursor and input text should have the same spacing as the placeholder
image
3. looks like the left padding on mobile is redundant 
image
4. this button and submit button on mobile should be fullwidth
image
5. this line usually appears between body and footer/header tags if body or html has light background, try to check this way or ask in fe_chat)
image

KseniyaGonchar:
thanks!

vlad-demchuk (changes requested):
do the same with form submit  and apply button
they should be fullwidth
image
image

AlisaVlasova (approved):

PR #531, opened 2022-12-04
commit message

11Olya112:


loralevitska (changes requested):
Keep up the good work! Let's improve
correct main photo and appearance of the mobile version. It should be on the right side and up to half of the left part

https://user-images.githubusercontent.com/70009205/206745875-f92e00c9-6bae-4b64-8871-56cc1a363793.mov
2. 
image
3. photos are compressed and text is gone
image

4. 
image
5.
image
6. add smooth scroll to page
7. add close icon to menu now you have a menu that opens when you click on the logo
image
8. Check all mobile mode, they are completely different and not only in the cap, but throughout the landing
image

polosanya (changes requested):
Hi! You did a great job, but for now the page looks unfinished.
Please check yourself with the design and try to improve it
1. Hover doesn't match the design
image
2. Add favicon to the page
image
3. Page is not centered and horizontal scroll is present
image
4. Problem with icons is not fixed
image
5. It doesn't look right
image
6. On mobile version there are also some problems:
Please check yourself with the design
image
image
image

Anastasiia-Svintsova (changes requested):
- The carousel doesn't match the design. It should be full width on a small screen
image
- Would be better to change it to the close button `X`
image
- Please, add a hover effect to all icons here 

https://user-images.githubusercontent.com/99403198/208236262-e0de7f62-d210-4ff6-b3f2-27d6d1ea6041.mov

Anastasiia-Svintsova (approved):
Great! Just remove this white background before adding it to your portfolio
image

PR #539, opened 2022-12-14
Dia landing

bojkovladislav:
[DEMO LINK](https://bojkovladislav.github.io/layout_dia/)


SanyaBratashchuk (changes requested):
1. shouldn't be a link 
image
2. Should be links to the real socials 
image
3. Change color of these links on hover 
image

PR #541, opened 2022-12-20
Develop

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/layout_dia/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -48,32 +48,32 @@
               id="togler"
             >
 
-            <ul class="nav__list nav__list-header">
-              <li class="nav__item">
+            <ul class="nav__list nav__list--header">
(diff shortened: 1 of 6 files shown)


Anna-Tkachenko (changes requested):
Good job! 
Some comments:
1) Will be good to add some styles on hover state for slider arrows
2) Social icons near the contact form should be the links on corresponding social networks (don't forget to open them in the new tab)
3) I can't submit the form, it redirects me to 405 error page, check it
4) Check the position of the button for mobile menu, looks like element are not centeredIt
image
5) Is it expected displaying?
image
6) Check indents between the form and screen side (the left one), probably some paddings were missed
image
7) Same here about paddings, check it
image

PR #546, opened 2022-12-28
Header was added

kostiantin-hudzenko:
[DEMO LINK](https://kostiantin-hudzenko.github.io/layout_dia/).


witflash (changes requested):
As I see you just start your work.
You can ask for a review when you will finish the whole landing page.

IvanFesenko (changes requested):
1st section should be 100% height on any screeen


This buttom should have ssame approach as same button on header section
2023-02-14_13h34_12

Fix autocomplete styles 
2023-02-14_13h35_11
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Form submit leads to 404

Why instagram and twitter icon lead to facebook?

IvanFesenko (changes requested):
Well done. 
Let's disable scroll in case when menu is open
To disable page scrolling under the menu add the next code
```
.page__body--with-menu {
  overflow: hidden;
}
```
```
window.addEventListener('hashchange', () => {
  if (window.location.hash === '#menu') {
    document.body.classList.add('page__body--with-menu');
  } else {
    document.body.classList.remove('page__body--with-menu');
  }
});
```

IvanFesenko (approved):
Congrats

PR #549, opened 2023-01-04
AIR 

the-tique:
[DEMO LINK](https://the-tique.github.io/layout_dia/).


SanyaBratashchuk (changes requested):
1. All external links should open in the separate tab

SanyaBratashchuk (approved):

PR #557, opened 2023-01-13
Develop

YevheniiHura:
https://YevheniiHura.github.io/layout_dia/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -308,23 +308,9 @@
                       Design Packaging
                     </p>
 
-                    <div class="services__card-container">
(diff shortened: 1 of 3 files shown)


IvanFesenko (changes requested):
Not all requirements were done:

- When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded
- The "Learn more" button should lead to the closest block (Our expertise)
-  Do not use 2 a tags for 'Learn More'

IvanFesenko (approved):
Good job

PR #568, opened 2023-02-03
finding it difficult to get the menu spacing the same as figma

spencerspe:


polosanya (changes requested):
Please add a demo link to your PR description, so we can properly check it)
image

polosanya (changes requested):
Keep up the good work!
1. Let's add a favicon
image
2. Font doesn't match the design
image
3. Add increasing hover effects to all the images
image
4. Same with cards. Also, "Learn more" should be a link
image
5. 404 error on burger menu icon click
image
6. Fix this problem
image

PR #571, opened 2023-02-05
solved

Olek-Lytvynenko:
[DEMO LINK](https://Olek-Lytvynenko.github.io/layout_dia/)


DarkMistyRoom (changes requested):
- check the font styles for this link. it looks too small
image

- check the background color in this section
image

- fix the form on the tab
image

- this indent should be bigger on the tab
image

- here should be some top indent
image

DarkMistyRoom (changes requested):
- check this section on the mockup. it looks wider
image

- it should be 20px
image

- these indents should be equal
image
image

PR #572, opened 2023-02-06
Dia_air problem with hover_before

p0mk:
https://p0mk.github.io/layout_dia/


loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

- It is worth limiting the width of the content for large screens (2k or 3k pixels, for example) 
image
image
- See how your page works on all screen sizes and fix errors
image
- 
image
- Open map in new tab 
image
- These blocks should be centered
image
- Fix icon hover 

https://user-images.githubusercontent.com/70009205/217022857-8d30901e-8f92-4af8-a3bb-0f4e6e16a239.mov

loralevitska (approved):
🔥

PR #585, opened 2023-02-16
Air(Dia)

BeeGooDee:
https://BeeGooDee.github.io/layout_dia/


DarkMistyRoom (changes requested):
- make the header fit screen height
image

- remove the background on the autocompleting
image

- something goes wrong
image

- check side indents on the tab/mobile. they should be smaller
image

- remove this line
image

ivbutsykin (approved):
Good job!

PR #591, opened 2023-02-24
dia-page

valeriy-mikheyev:


nazarmatsevych (changes requested):
Good work, but:
Mobile view does not match the design
image
image

slavanosatskyi (approved):
Very nice work 😎

PR #595, opened 2023-03-03
added header for check

Viktoria-Moroz:
[DEMO LINK](https://Viktoria-Moroz.github.io/layout_dia/)


anstsot (changes requested):
Good start, re-request the review once you finish the whole page. Good luck with the task!

DarkMistyRoom (changes requested):
Look well. I like your hover effects. Few moments:
- would be prettier to make header fit the window height
image

- cursor doesn't become a pointer over this button
image

- add hover effect to these links too
image

- these should be links to real socials
image

- remove background on the auto-completing
image

- as we don't have menu in the mockup, it's up to you, but I'd prefer to leave logo  and make the top padding bigger
image

- these cards have equal width on mobile according to mockup
image

SerhiiKirik (approved):
Great 🔥 

Just change only the border-bottom on hover, bc you have items jump on hover
image

too big right gap, mb would be better make place-content: end
image
end
image

PR #596, opened 2023-03-03
dia for review1

mzaloha:
https://mzaloha.github.io/layout_dia/


etojeDenys (changes requested):
almost perfect


nit:  The "Learn more" button should lead to the closest block (Our expertise)

Anastasiia-Svintsova (changes requested):
- Would be great to remove styles for autofill
image

- Here is the issue with the carousel image on the small screens 
image

- Good job! But I can't approve it for now. Fix the issue with the carousel, please

mzaloha:
@Anastasiia-Svintsova sorry, I didn't get it. what is the error with carousel?
and the second question about autofill. should it be done with js? thank you!
image

etojeDenys (approved):
LGTM

PR #610, opened 2023-03-13
Project2

Ryshon:
https://Ryshon.github.io/layout_dia/)


witflash (changes requested):
Good work.
Here are some points.

1. You have to add HTML validation for your form: 
image

2. Your `about us` anchors don't lead to the correct section:
image

3. There is no padding for the text section, but it should be.
image

4. Social icons should be white by default and blue on hover:
image
image

PR #613, opened 2023-03-17
air page

myplekan:
https://myplekan.github.io/layout_dia/


slavanosatskyi (changes requested):
nice work, let's polish it a bit
1. Add hover effect to all these elements (except icons, they work fine)
Screenshot 2023-03-17 at 17 41 45
Screenshot 2023-03-17 at 17 42 15
2. Menu click doesn't work on small screens
Screenshot 2023-03-17 at 17 43 31

loralevitska (changes requested):
Almost perfect) Let's improve

- All links should be open in new tab (socials, map, etc...)
- fix placeholder `email@example`
- Menu click still doesn't work. A side menu should open instead of scrolling to the bottom of the page

 If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

OleziO (changes requested):
Prevent reloading of the page after submission.

Fix menu. It must be full-screen and have **X** button.
image

PR #618, opened 2023-03-22
Dia

Ryshon:
https://Ryshon.github.io/layout_dia/


etojeDenys (changes requested):
1. I can't submit form or fields do not reset after submitting
2. rewrite it to be more responsible
image
3. the “Apply” button also should lead to the contact form

PR #628, opened 2023-04-05
Dia / Air landing

den-stronin:
[DEMO LINK](https://den-stronin.github.io/layout_dia/)


VitaliyBondarenko1982 (changes requested):
Good job.

Needs some improvements
- add cursor pointer for all interactive elements (buttons, links)
- make header background on all width (as footer)
Screenshot from 2023-04-05 12-04-17
- consider to make autocomplete styles as default background for form input
Screenshot from 2023-04-05 12-05-15

den-stronin:
1. Added cursor pointer for "contact__button button". 
2. Removed "max-width: 1600px" for the header. It has the full width of the page, but I can make "max-width: 2560px" for header and footer.
3. Added default background for form input.

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻 
https://www.loom.com/share/2cfb7b6d1dad4e20bbbc327aa6a4b8ed

den-stronin:
1. Margin and padding property values changed to multiples of 4 and 8.
    2. "Transition: all" are replaced by "transition: color", "transition: transform" and so on.
    3. The "Menu close button" does not disappear from the menu page when the screen size and orientation change. The "Menu open button" disappears in the header (transform: translateY(-100px)) if the screen width is greater than 1023px;
    4. Added "target_blank" to social media links
    5. The "transform: scale(1.1)" property on hover is disabled for mobile devices, but saved for devices with a screen size of more than 1023px.

mykhalenych (approved):

PR #631, opened 2023-04-07
Solution

undepend:
[DEMO LINK](https://undepend.github.io/layout_dia/)


TarasHoliuk:
1. Typo (secnion)
image
2. Background color doesn't match the mockup:
image
image
3. There is an example of hovered button style in the mockup (color should be still blue, but a bit darker):
image
  Your app:
  
image
4. Autocomplete styles look bad here:
image
Should be disabled ([link](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill)) 
5. Keep the padding for the menu the same as for the header:
image
image
6. On screens with a width < 640px the button is full-width (as in the mobile version), but the form isn't - be consistent here (not critical, but it would be nice to fix it)
image
7. Add favicon
8. Scrolling should be smooth (after clicking on an anchor link)

TarasHoliuk (changes requested):
Indeed, very good job :+1: 
But let's improve it a bit (check comments)

tetiana-ivanchenko (changes requested):
autocomplete styles still ddin't fix

undepend:
> Indeed, very good job 👍 But let's improve it a bit (check comments)



> autocomplete styles still ddin't fix

I tryed to fix it. I don't know whether it worked, because I can't see what is wrong (in my chrome autofill doesn't work).

PR #634, opened 2023-04-10
layout Dia

misha-vynnyk:
[DEMO LINK](https://misha-vynnyk.github.io/layout_dia/).

I wrote this code myself without watching Misha's video. Should I continue in the same spirit? Am I doing something wrong?


OleziO (changes requested):
Demo: 
image
Design: 
image

Check all other breakpoints.

PR #639, opened 2023-04-21
air version1

tetyana-shatrova:
(https://tetyana-shatrova.github.io/layout_dia/)


SanyaBratashchuk (changes requested):
1. Add hover effects to all images
2. According to the design. These images should be rounded 
image
3. Why you make it via image? 
image
4. Add hover effects for buttons 
image
  5. This form is too short according to the design 
image
6. Remove background on autocomplete 
image
7. Form submit shouldn't trigger page refresh

SanyaBratashchuk (changes requested):
1. Add favicon
2. Add transition for hover effects
3. Fix textarea 
image
4. Should be links to the real socials 
image
5. 
image
6. Change color of this links on hover 
image
7. All external link should be open in the separate tab 
Please check your work with checklist before re-request  review

tetyana-shatrova:
> Good job, but there are few points to fix
> 
> 1. add hover effect to the links in header, footer and to the button "Hire us"
>    image
> 2. change the placeholder
>    image
> 3. Fiw the border-radius on tablet
>    image
> 4. the menu isn't working on tablet
>    image

tetyana-shatrova:
In figma there were no menu on Tablet, so i didn't make it

loralevitska (changes requested):
Almost perfect)
Come up with a menu design, it can be arbitrary, now for the recruiter it will look as if the functionality does not work, because the menu icon is there, but nothing happens when you click
Also implement clicks on all logos (return to the top of the page)

loralevitska (changes requested):
Much better) A few points remain
- add cursor pointer to each clickable elements (links, buttons, socials)
- add smoothness to the appearance and disappearance of menu

tetiana-ivanchenko (changes requested):
1. fix autocomplete styles
image
2. 
image

tetyana-shatrova:
> 

Hi! Autocomplete is off.

OleziO (approved):

PR #641, opened 2023-04-26
AIR landing page

mttony:
DEMO LINK:

https://mttony.github.io/layout_dia/


DarkMistyRoom (changes requested):
- use the container to limit content width everywhere, including the header and top-action in the footer
- there should be fixed size of buttons on the desktop and tab screens
image

- add placeholders to the form
image

- bottom border disappears on focus
image

- I can't see what I typing. and remove the background on auto-completing
image

- add titles to this section
image
image

- use the same top-actions in the menu
image

- fix the side padding. it should be 41px on the tab
image

- fix the slider
image

- the cards in this section should be placed in the grid(tab)
image
image

- limit text width in this section(tab)
image

image

- remove this line
image

- this button should be 100% width on the mobile
image

tetiana-ivanchenko (changes requested):
1. remove this gap
image
2. it should be links on the social media
image

3.change colors
image
4. check all points from the checklist, and fix it
5. fix scroll
image

PR #645, opened 2023-05-07
Develop

denis-kononenko:
[DEMO LINK](https://denis-kononenko.github.io/layout_dia/)


vlad-demchuk (changes requested):
1. add some scale on hover
image

2. these should be links to social media
image

DarkMistyRoom (changes requested):
Good job! You're almost there
- add some hover effect to these cards too
image

- good job with the menu. But would be better to close it after clicking a link. It's more expected behavior
image

- limit text block width in these cards
image

- the form should fit the first 4 columns
image

nazarmatsevych (approved):
👍

PR #649, opened 2023-05-11
AIR_landing

uliana-kudriavtseva:
- [DEMO LINK](https://uliana-kudriavtseva.github.io/layout_dia/)


TarasHoliuk (approved):
Nothing to add :+1: 
That's the best work I've seen :fire: :open_mouth:

TarasHoliuk (approved):
Nothing to add :+1: 
That's the best work I've seen :fire: :open_mouth:

DarkMistyRoom (commented):
- would be great to make this corner white

image

- if you want, it would be nice to make the slider smooth. But it's already good
- the button on a mobile screen can be placed a bit higher for me because it's too close to the slider when there is enough space above the button
image

PR #652, opened 2023-05-14
My landing #2

mrsargos:
[DEMO LINK](https://mrsargos.github.io/layout_dia/).


VitaliyBondarenko1982 (changes requested):
Nice job.
Please proofread [checklist](https://github.com/mate-academy/layout_dia/blob/master/checklist.md) and fix some points
- white favicon not good visible
 
Screenshot from 2023-05-14 16-49-49
- All interactive elements(links, buttons, pictures, icons) should have a hover effect and cursor pointer. User must intuitively understand that he can interact with them
- The user must have the opportunity to conveniently write 2-3 lines of text in the message input field 
  
Screenshot from 2023-05-14 16-54-22
- When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded
Screenshot from 2023-05-14 16-48-03
- After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
  
Screenshot from 2023-05-14 16-46-24
- All the social icons in the footer should be clickable and open the social networks in a new tab
  add hover effect for all social icons. also add transition for smooth changing color of icon.

polosanya (changes requested):
Almost done!
1. It would be great to add some hover effect to the link too, not only icon.
Also, you can even add hover effect for the whole card
image
2. Form doesn't match the design. Pay attention to the 'message' input and 'Send' button
image
image
3. It would be great for main screen to take the whole height
image
4. Twitter icon doesn't change size on hover. Fix it or just change the color for all icons on hover
icons
5. It's not critical, but would be great to change the title, as 'Dia' is an old name 
image

mrsargos:
> 

image

I don't know why but with the button everything is ok

volodymyr-soltys97 (changes requested):
In general, it is good, but some points need to be corrected

This is not done by design
image
You need to look at that white line on the phone
image

vlad-demchuk (approved):

PR #653, opened 2023-05-15
Develop

VolodymyrPst:


volodymyr-soltys97 (changes requested):
Good start, but let's fix some bugs
These links don't take us anywhere
image
Let's take the border and apply other styles when hovering
image
You should add styles for these buttons when on hover
image
Here the text is not centered
image

PR #655, opened 2023-05-16
Develop

VolodymyrPst:


volodymyr-soltys97 (changes requested):
Great start 🚀 
Let's make these links in the footer take us to the relevant sections, and not just to the top of the page
image
Here you should add links to these social networks and open them in a new tab
image
It is also worth adding the transition-duration property everywhere, where you change the styles when hovering for a better look

PR #657, opened 2023-05-17
Develop

VolodymyrPst:


YegorKorenkov (changes requested):
1. Please remove scroll, when the menu is open
image
2. The "Learn more" button should lead to the closest block (Our expertise)
3. The "Hire Us" button in the header menu should lead to the contact form
4. In the Vision, Passion, Results section, the “Apply” button also should lead to the contact form
5. Please check paddings for mobile screens, also text should be align left
image
image
6. Please fix this
image

PR #660, opened 2023-05-20
Develop

VolodymyrPst:


vlad-demchuk (changes requested):
1. header doesn't match the mockup
slider should be in the lower right corner for any screen and header should be 100vh
image

image

2. align text inside inputs and fix styles for autocomplete
image

3. fix menu for 320px width
image
and this 
image

PR #665, opened 2023-05-23
Develop

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/layout_dia/


polosanya (changes requested):
Nice! Let's improve your page a bit)
1. Make sure that you add transition for the default state of the element, not hover, so animation works smoothly both ways
animation
2. This hover effect doesn't look correctly too
hover-link
3. Same here. Also, make sure that transition duration isn't too big and it is the same per page.
0.3s is the best choice for user experience 
image
4. Email placeholder is wrong. It should be 'Your email'
image
After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
5. All external links should be opened in a new tab. Also, check their transition too
image
6. It would be great to implement simple burger menu
image

polosanya (approved):
Looks great! Nice!

PR #667, opened 2023-05-24
Develop

VolodymyrPst:


vlad-demchuk (changes requested):
1. slider bigger than header
image
2. make all fields required and change autocomplete styles
image

PR #670, opened 2023-05-25
Develop

VolodymyrPst:


volodymyr-soltys97 (changes requested):
Good job 👍 
But let's fix some mistakes 
image
Add animation for this element
image
Need to fix
image
image

PR #673, opened 2023-05-29
task done

andriy-litkovets:
[DEMO LINK](https://andriy-litkovets.github.io/layout_dia/)


src/blocks/about-us.scss
```diff
@@ -0,0 +1,49 @@
...
+  &__title {
+    font-family: "Poppins", sans-serif;
+    font-weight: 600;
+    font-size: 32px;
+    line-height: 48px;
+    color: #253757;
+  }
+
+  &__subtitle {
+    font-family: "Open Sans", sans-serif;
```
OleziO, line 37: Move to the html selector.

src/blocks/agency.scss
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @include OnSmallDesktop {
+      margin-bottom: 84px;
+    }
+  }
+
+  &__button {
+    display: block;
+    box-sizing: border-box;
```
OleziO, line 52: Move to the * selector and remove it from all other places.

OleziO (changes requested):
Fix the menu, it must be a separate frame (prevent scroll).
image

sTorba24 (changes requested):
Almost good! Lets improve a bit

1.carousel seems broken a little on tablet
image

2.icons should be center-aligned 
image

3. Add hover effects on images and cards

4.Reset form on submit

5. It should take all width of page on table
 the text should be wider. Also check font-weight
image
mockup:
image

6.buttons should change color on hover

7. hover effect on icons
mockup:
image


your:
image

witflash (changes requested):
Let's make a few improvements.

1. Could you add HTML validation for your form? There is no validation for now
image

2. It's better to make a not-white favicon. Because on the light tab it looks invisible :)
image

3. Could you test your Slider on page initialization? Because it doesn't work on init, only after I've changed the page width.
image

OleziO (approved):

PR #683, opened 2023-06-08
Header with slider

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/layout_dia/).

Pleace, check my slider code


src/scripts/main.js
```diff
@@ -1 +1,24 @@
...
+
+document.querySelector('.slider__button-left')
+  .addEventListener('click', function() {
+    offset = offset - 33.333333;
+
+    if (offset < 0) {
+      offset = 66.6666666;
+    }
+    sliderLine.style.transform = 'translateX(-' + offset + '%)';
+  });
```
anastasiiavorobiova, line 24: There is no error handling at all. If there is nothing in the DOM, it would break.

anastasiiavorobiova (changes requested):
Really good-looking slider!
* Consider adding checks to manage cases where there are no such DOM element
* It's bad practice to access DOM elements by class. Consider using ids or data attributes to find needed DOM elements
* There are a lot of magic strings in your javascript consider using variables with meaningful names
* Consider adding functionality to handle more than 3 pictures in the slider. For this project, it may be unnecessary but in real-life projects, we never know how much content would be there

vovaAlieksieiev:
Hi. I fixed the slider code a bit. also added a menu block and a footer. I look forward to comments)

demetrxx (changes requested):
Fix demo link

PR #710, opened 2023-07-09
Develop

Zaransoftline:
https://Zaransoftline.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
You need to add additional styles for all clickable elements when hovering

anastasiiavorobiova (changes requested):
Awesome work! Unfortunately, navigation and buttons don't have a hover state, consider fixing it

volodymyr-soltys97 (changes requested):
1. You need to implement burger menu when I click on this icon
image

2. Add additional styles for these clickable elements when hovering
image
image

volodymyr-soltys97 (changes requested):
Need to remove scroll when this menu is open 
image

PR #714, opened 2023-07-12
Solution

nugardesq:
https://nugardesq.github.io/layout_dia/


lerastarynets (changes requested):
Pls watch the video review :)

https://github.com/mate-academy/layout_dia/assets/77551820/d2c0ba39-a04f-408b-a380-9330800875cd

polosanya (changes requested):
Keep up the good work!
1. These cards don't match the design. Double-check the sizes
image
image
2. Learn more links should have a hover effect and be clickable too
learn-more
3. This button does nothing on click
image
4. Form doesn't match the design
image
image
5. After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
image
6. There is still an error on the form submit
image
7. Add hover effects to icons
image

lerastarynets (changes requested):
The issue with autofill is not fixed, also for some reason you have double border for inputs
Screenshot at Jul 24 16-58-13

Anastasiia-Svintsova (changes requested):
Keep going! Check the design more thoroughly 🙏 
- The carousel doesn't look correct. Fix it according to the design, please 
image

- Title doesn't match the design as well 
image
image

- Check the space between the articles on large screens 
image

- The same goes for all sections. You need to add a max-width for the content
image

- Styles after auto-filling doesn't look good. It should look like an average input  
image

- All external links should be opened in a new tab
- The user should be able to open the menu
- No underlining here 
image

- Add correct bottom padding here 
image

- Check spaces here 
image

agniya-i (approved):
I approved, but pay attention to several moments 

1. Check margin between subtitle and card on the mobile/tablet, and subtitle font-weight
Screenshot 2023-07-28 at 19 17 34

2. It would be better to make autocomplete text color white
Screenshot 2023-07-28 at 19 22 25

PR #715, opened 2023-07-13
add task solution

Daryna-Kukharets:
[DEMO LINK](https://daryna-kukharets.github.io/layout_dia/)


etojeDenys (changes requested):
1. let's add some hover to these btns
image



2. let's disable scrolling when the menu is open
image

PR #722, opened 2023-07-20
Develop

mykola-kharin:
[DEMO LINK](https://mykola-kharin.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good start 👍 
It is necessary to improve the logic of this menu when it opens, so that when we click on the icon to open it, we are not thrown down the page
image
Keep doing

Anastasiia-Svintsova (changes requested):
- The menu should have a full height 
image

- Please, implement the whole page before requesting the review again

PR #727, opened 2023-07-27
Dia

Dimoshua:
[DEMO LINK](https://Dimoshua.github.io/layout_dia/)


loralevitska (changes requested):

- Make sure that all interactive elements have a cursor pointer on hover
- continue placeholder 
image
- disable default autocomplete background color, it breaks design
image
- fix this title according design for all modes 
should be
image
yours variant 
image

loralevitska (approved):
Well done)

- add cursor pointer to this buttons 
image

PR #735, opened 2023-08-04
add solution

OleksandraKoshyk:
[DEMO LINK](https://OleksandraKoshyk.github.io/layout_dia/)


anastasiia-tilikina (changes requested):
Great job! and nice hover effects! 🔥
But here we have things to improve:

1.  when I press slider buttons page automatically scrolls to top. It looks a bit glitchy though) let's fix it.
image

2. add some hover effects for nav in header and footer.
image

3. Something bad is happening with layout here) slider overlaps button
image

4. When I hover over the entire card hover effect works for arrow button as well. lets add it only when I hover over button.
image

5. I'd suggested you to make these links opened in new tab. And you can go even further - find real account in social media of this company) but it is not mandatory though.
image

6. I see some white space in the bottom of the page.
image

7. The submit button doesn't have cursor pointer
image

8. when menu is opened on mobile scroll should be disabled. Now it is glitchy, because some content overlap menu
image

9. I'd suggested you to add some smooth appearance of mobile menu.
10. To add even more interactivity on your landing page you can make these links as well.  the address will lead to google maps.
image

etojeDenys (changes requested):
only one nit:
fix these autocomplete styles
image

VitaliyBondarenko1982 (changes requested):
Great job!
Almost done.
Just one more nit)

Consider to make custom scrollbar
https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp
Screenshot from 2023-08-07 13-06-58

etojeDenys (approved):
lgtm

PR #736, opened 2023-08-04
implementation

NadieinOleh:


maxim2310 (changes requested):
Please attached demo link to PR

Anastasiia-Svintsova (changes requested):
Nice start! Check the design more thoroughly 

- Check the text position on the desktop. It doesn't match the design 
image

- Where are the buttons? 
image

- Nav links don't work. And add hover effects to them
image

- There shouldn't be a menu icon on the Desktop 
image

- Use SVG for the logo for better image quality 
- The button should be rounded 
image

- This hover effect looks strange in my opinion. Could you use a better approach to avoid overlapping?
image

- Cards don't match the design. And add a hover effect as well 
image

- This button should be rounded, has a hover effect, and lead to some section.
image

- All external links should be opened in a new tab 
- There is no placeholder for the message 
image

- Remove styles for autocomplete 
image

- The form doesn't work correctly. It should be cleared after submitting 
- Add hover effects to these links 
image
- The logo in the footer should lead to the top of the page

maxim2310 (changes requested):
Nice work, but still have some bugs:
1.  in my screen i dont see gallery button at all 
image
2. need to use same hover effects 
image
3. this block is differ from design in mobile
image
4. menu button is not working 
5. on small mobile gallery is hidden a bit
image

maxim2310 (changes requested):
Almost there) Just ones more thing needs to be fixed
1. 
image
2. still have a different hover on it 
image
3. menu button is not working

witflash (changes requested):
Good work!

There are several things to improve.

1. It's better to add cursor: pointer on hover any interactive elements. For example, buttons on your slider:
image
It's the best practice for user experience.

2. I see you added a blue color on insta hover, but it should be the same behavior for any other social icons. It should be a general solution.
image

3. It may be something went wrong, but the mobile menu doesn't open on the menu click.
image
Could you check it, please?

Other things look good for me. Good job!

sTorba24 (changes requested):
Almost good

1. Add hover effect on links(phone number, map) Also, remove target _blank from number
image

2. image should be sticked to right side 
image

now
image

3. Add hover effects on card or card elements 
image

4.should be cursor pointer here
image

5.Menu still not working on mobile, fix it please

lerastarynets (changes requested):
Did you forget to redeploy the project?) Previous comments aren’t fixed

witflash (approved):
Good job!

You can improve a little bit and remove scroll for the whole page if your mobile menu is opened:
image

You can use `overflow: hidden` for the `body` if menu is opened and remove this rule is menu is closed.

PR #739, opened 2023-08-10
Hello, can you help me with slide?

PavlovaOlena:
[DEMO LINK](https://PavlovaOlena.github.io/layout_dia/)


polosanya (changes requested):
Nice! Let's improve your page a bit)
1. It would be great to add hover effects to all interactive elements
image
2. Same with images. Let's increase them on hover
image
3. After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/).
Also, it would be great to reset the form on submit
image
4. Main image is missing
image

witflash (approved):
Good job!

About slider: at first you can study the JS modules and then you can come back to this work and extend your solution with your new knowledge of JS. It's ok to decompose the solution of the whole landing page.

PR #745, opened 2023-08-16
add task solution

ydem1:
https://ydem1.github.io/layout_dia/


xiimiik (changes requested):
there should be one thing, either sharp corners or rounded edges
image
must be correct font
image
hovers here
image
image
must be with cursor pointer
image
i cant hide menu on one more button click
image

loralevitska (approved):
Good job! Check my comments for improvement 

- remove default autocomplete background-color, it breaks design
image

PR #758, opened 2023-08-29
add task solution

ViraMMM:
[DEMO LINK](https://ViraMMM.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Add hover effects for these elements
image

2. When I click phone number I get this problem
image

3. This link should be opened in the new tab
image

volodymyr-soltys97 (changes requested):
The content should not stretch across the entire width of the window, but should be in a container that will have a fixed width and will be centered on the page, check how this is implemented in the design
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review

volodymyr-soltys97 (changes requested):
Some sections are evenly stretched across the entire width, this needs to be fixed

volodymyr-soltys97 (changes requested):
1. These links should lead to the relevant sections
image

2. Check the left and right indents on the mobile version
image

3. Need to implement burger menu, when I click on this icon
image

volodymyr-soltys97 (changes requested):
1. The content in the header section must also be placed in a limited-width container, as this is implemented in other sections
image

2. Still not fixed the left and right indents in other sections, need to be fixed everywhere on the mobile version
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)
have error page on click 
image

volodymyr-soltys97 (changes requested):
1. Add `cursor: pointer` and hover effects for all clickable elements
image

2. These links should lead to the relevant social networks
image

volodymyr-soltys97 (changes requested):
1. These links should be opened in the new tab
image

2. Need to fix the indents between the menu and next section
Demo
image

Design
image

3. Need to center it horizontally
Demo
image

Design
image

anastasiiavorobiova (approved):
Looks awesome!

PR #760, opened 2023-08-30
task solution

no4kar:
https://no4kar.github.io/layout_dia/

it is raw


DarkMistyRoom (changes requested):
Good job! Keep working and check the comments

- add a smooth scroll to the page. When I click on a link in the header it should navigate smoothly to the appropriate section
- add some hover effects to this section
image

- would be good to add possibility to swipe the slider by finger (tab/mobile modes)

maxim2310 (changes requested):
Good job! Let's improve your code a bit)
1. remove autocomplete style 
image
2. add hover effect here
image
3. add fav icon

mykhalenych (approved):

PR #761, opened 2023-09-02
Dia landing

kostyaniekrasov:
[DEMO LINK](https://kostyaniekrasov.github.io/layout_dia/).


anastasiia-tilikina (changes requested):
well done!
let's just add small fixes

1. this pic should be full height
image

2. let's add transition on buttons for smooth color changing on hover
image

3. add hover effects on img only, not on entire block
image

4. these blocks should be full width on mobile
image

5. something is wrong with transition here. block overlaps the previous one
image

lerastarynets (changes requested):
Pls watch the video review :)

https://github.com/mate-academy/layout_dia/assets/77551820/270596f9-4c72-4959-80de-332949e9603b

VitaliyBondarenko1982 (changes requested):
Good job.
Please fix some moments.
- add triansition for elements with hover effects. and for all propties which change.
  and pay attention that for make transition in two ways you need to all transition for main selector instead for selector with hover.
  
Screenshot from 2023-09-04 18-17-34
chrome-capture-2023-8-4 (2)
- fix autocomplete styles
Screenshot from 2023-09-04 18-18-50
- suggest instead of setting overflow hidden for textarea make custom scrollbar
https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp
Screenshot from 2023-09-04 18-18-39

maxim2310 (approved):
Good work!!!

PR #767, opened 2023-09-08
Add the first version of air-layout

PodvAx:
[DEMO LINK](https://PodvAx.github.io/layout_dia/)


etojeDenys (changes requested):
awesome animations. few improvements:
1. the menu looks strange let's stretch on full-width

https://github.com/mate-academy/layout_dia/assets/70220667/0ddca228-8e3c-478c-bb87-737bbaa7f5cb


2. reset auto-complete styles 
image

sTorba24 (changes requested):
Looks really cool!


check submit behavior and prevent it, because now it's reloading page and changing url

PodvAx:
I don't know why, but on my local host, all work nice. But when i deployed it, on mobile phone, after submitting the form, it don't want scroll to the top. It scroll to the top of the page, only if there is no anchor in url. I already code, that url cleans on load page, but all in vain. And some browsers may finish with 405 error when submitting the form, although in Safari on my phone and in Chrome all works nice.

DarkMistyRoom (changes requested):
- don't reload the page on form submitting, just clear the form. check 15 option [here](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

etojeDenys (approved):
lgtm

PR #768, opened 2023-09-12
solution

Spec0s:
[layout_dia](https://spec0s.github.io/layout_dia/)


anastasiia-tilikina (changes requested):
Well done!
Yet we have some issues to fix:

1. fix padding in header and footer
image

image

2. Let's add hover effects on for img in not for entire block here
image

3. google maps link should be opened in new tab

4. these cards should be the same height
image

5. lets place nav links a bit lower.
image

6. form submitton shouldn't reload page. it should only reset fields.

Spec0s:
i did not find problem with footer ?

anastasiia-tilikina (changes requested):
footer in figma:
image

your footer:
image

image

same thing with header:
image

image


https://github.com/mate-academy/layout_dia/assets/108538921/517eebff-e269-4d34-9ea1-e5bff55c514e

anastasiia-tilikina (approved):
GJ

PR #773, opened 2023-09-21
Dia page

ikispower:
[DEMO LINK](https://ikispower.github.io/layout_dia/)


nazarmatsevych (changes requested):
1. Clear form fields on button click, also page should not reload on submit

2. This links are not working, also add some hover effects 
image

3. Menu button not working
image

volodymyr-soltys97 (changes requested):
1. Still not fixed from the previous review
image

2. Need to add cursor pointer for the button in the form
image

3. Still not fixed from the previous review, you need to implement the burger menu 
image

sTorba24 (changes requested):
1.Add validation to form, also clear it after submit
image

2. margin between title and elements is too big(check mock)
image
image

nazarmatsevych (approved):
👍

PR #777, opened 2023-09-25
new task

Vovchukkul:
https://Vovchukkul.github.io/layout_dia/


etojeDenys (changes requested):
1. should be the button with some hover and cursor: pointer
image



2. add some hovers
image
image


3. should be links
image


4. should be the form button
image


5. disable scroll when menu is open
6. add favicon

src/index.html
```diff
@@ -1,13 +1,395 @@
...
+      rel="icon"
+      type="image/x-icon"
+      href="./images/sub-title/sub-title-icon-dark.svg"
+    >
   </head>
   <body>
-    <h1>Dia</h1>
+    <header class="header">
+      <div class="top-bar">
+        <wrap class="wrap">
```
etojeDenys, line 26: still not fixed comment with wrap

etojeDenys (changes requested):
1. you should add some container to fix strange behaviour on big screens
image


2. When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded
3. The form shouldn’t submit empty
4. When a user clicks on Address Google Maps is opened in a new tab
5. In the PROCESS section add a hover effect to cards (for example, let them increase in size a little)

src/index.html
etojeDenys: image

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1. need to remove autocomplete styles 
image
2. name and message must be required 
3. would be nice to add hover to these cards also
image

src/styles/utils/_mixins.scss
```diff
@@ -4,3 +4,79 @@
...
+    padding-inline: 54px;
+  }
+
+  @include onBiggestDesk {
+    padding-inline: 227px;
+  }
+
+  @include onLargestScreen {
+    padding-inline: 707px;
+  }
```
sTorba24, line 25: it's really not good idea :(

sTorba24 (changes requested):
1. this section isn't centered
image

2. set max-width for content inside blocks(you have mock for big screens and content still has such width)
image


3. why text is black inside input? it is very inconspicuous
image

4.remove target _blank from `call` link 
image

etojeDenys (changes requested):
1. looks broken
image
image



2. these links should lead to some sections
image

etojeDenys (changes requested):
try to re-deploy your site, cause nothing changed

etojeDenys (changes requested):
still can reproduce the comments, feel fee to ask for some help in chat

Vovchukkul:
> 1. looks broken
> 
> image image
> 2. these links should lead to some sections
> 
> image

can you explain what the problem is? because I think that the problem is in distance between send button and air block

Vovchukkul:
this is how it looks like in my latest pr
image

etojeDenys (approved):
good job

PR #780, opened 2023-09-29
Develop landing_dia

boytrend0108:
https://boytrend0108.github.io/layout_dia/


sTorba24 (changes requested):
Good work!

1. align links with logo vertically

image

2. when i click right it scrolling left, and opposite when i click left
image

3.prevent default behavior when form submitted and clear it
image

4. make background full screen width but items inside restricted max-width
image
(you need additional container)

5.add hover effect on links
image

6. fix mobile menu(remove y-scroll, and menu should cover whole page)
image

sTorba24 (approved):
I approve it. But pay attention to fix last bug 

when i zoom out, slider crawls out of the screen and x-scroll is visible
image

PR #787, opened 2023-10-10
landing has been done

MyNovikov:
https://mynovikov.github.io/layout_dia/


witflash (changes requested):
Hi!
Nice start! Keep going on your work!

Here are recommendations to fix current work.

1. Check this block and create a limited width for the text.
Current:
image
Reference:
image

2. Could you use better quality for this image? Now it looks very blurred
image

3. Check the font style and padding/margin for the header for this block
Current:
image
Reference:
image

4. Add some hover effects for this block (you can do it to your own taste)
image
It happens nothing when a user moves a mouse cursor over the block. And it's not a good practice.

5. Try to center all blocks on the page (the same as for our expertise block). I think you can use some containers with a limited width and auto margin-left and margin-right.
image

6. Check the styles for the form. It looks different from the mockup.
Current:
image
Reference:
image

7. On mobile the control buttons are placed vertically but they should be placed horizontally.
image

OleziO (changes requested):
Take away click styles after it has been made and the cursor is moved to the side
image
Demo: 
image
Design: 
image
Demo:
image
Design:
image
Demo:
image
Design:
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
Needs a couple improvements.
- add smooth scrolling to proper section after cliick on nav link
- [fix default autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/) for inputs
Screenshot from 2023-10-21 17-25-24

anastasiia-tilikina (changes requested):
Let's add some more improvements to your landing to make it more attractive)

watch the [video](https://www.loom.com/share/9e6c3749126546a58a4bb2c827968d83?sid=6bbda21b-5aec-4877-91b0-5b2cef1f7f1c)

IvanFesenko (changes requested):
You done a greate work. Lats item left - disable scroll in case menu is open

VitaliyBondarenko1982 (approved):
Well done

PR #789, opened 2023-10-11
Layout_dia

Deemacr:
[DEMO LINK](https://deemacr.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The user will not be able to change the size of this field
image

2. Add hover effects for all clickable elements and buttons
image

3. You need to correct the positioning of this image on the right as it is on the design
image

etojeDenys (changes requested):
1. remove outline and autocomplete styles 
image


2. fix the error on logo click
3. add hovers to all clickable elements 
image


4. The button should be stretched
image

5. I cannot open the menu on the burger button click
6. let's stretch on mobile the form
image

etojeDenys (approved):

PR #793, opened 2023-10-17
add_layout_air

JokerMTM:
[DEMO LINK](https://JokerMTM.github.io/layout_dia/).


etojeDenys (changes requested):
1. you need to stretch this photo
image

2. remove autocomplete styles
image

xiimiik (changes requested):
broken line on click (tablet resolution)
image

gap is such bigger as in design
image

double check margins here
image

text in buttons is not centered
image
image

strange empty space here, check design
image

Andriy-Shymkiv (approved):

PR #806, opened 2023-11-03
task solution

Krzychu92:
I don't know how to manage to set this two squares behind a photo. May anyone help me out?

image


volodymyr-soltys97 (changes requested):
Hello, for these blocks you can use 2 pseudo-elements before and after, and write z-index for the image

Krzychu92:
> Hello, for these blocks you can use 2 pseudo-elements before and after, and write z-index for the image

Okay I have changed it but how about position? Od somehow possibile to read from figma how to top/left this element or just try and compare to figma?

PR #807, opened 2023-11-05
DIA

ruslanagnatenko:
[DEMO LINK](https://ruslanagnatenko.github.io/layout_dia/)


polosanya (changes requested):
Nice start! Here are some comments about how you can improve your page:
1. Let's add some action for all interactive elements on click. (e.g navigate user to some section)
image
2. Same with navigation links. They should work correctly
image
3. Add increasing hover effects for all images
image
4. Pay attention to the font styles here. Also, let's add some hover effects
image
image
5. After autocomplete the form, change the [default styles](http://joxi.ru/EA4xgG8HXwL4lA). Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
 Make all input fields required 
image
6. Phone and address links should be clickable and work correct
image
7. Same with social media links. They should open some social page in a new tab
image

[Additional] Would be great to implement the slider
image

PR #808, opened 2023-11-07
Develop

Dmytro-Kretov:
[DEMO LINK](https://dmytro-kretov.github.io/layout_dia/).


loralevitska (changes requested):

- make `hire us ` text on the same level with nav links. The text should be centered and blue line underlined like absolute element from  `hire us` text
image
- fix nav and logo centering according design 
image
your solution 
image
- try to implement slider
- 
image
- disable text-area resize, it breaks design
image
- make all form field required 
- stretch content and slider to full width according design (mobile and tablet mode)
image
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
Fix a couple moments.
- add cursor pointer for send button
- fix horizontal scroll. it exists almost on any widht of screen.
Screenshot from 2023-11-09 12-24-15

VitaliyBondarenko1982 (approved):
Well done

PR #826, opened 2023-11-30
add task solution

TetianaBukhenko:
[DEMO LINK](https://tetianabukhenko.github.io/layout_dia/).


anastasiiavorobiova (changes requested):
Great start! 
Screenshot 2023-11-30 at 14 28 55
Consider fixing validation errors 
Screenshot 2023-11-30 at 14 26 28
When the menu is open, parts of the page content cover the menu
Screenshot 2023-11-30 at 14 25 43
Check the layout here 
Screenshot 2023-11-30 at 14 25 06
Screenshot 2023-11-30 at 14 25 11
Check an alignment here 
Screenshot 2023-11-30 at 14 23 48
These links should lead to the new tabs (Google Maps, the first page of proper social network)

ivbutsykin (approved):
Well done!

I found some moments that you can quickly fix:

1. Double check button in "Send us a message", it's doesn't have "cursor pointer" effect:
Screenshot 2023-12-06 at 20 34 19

> All interactive elements(links, buttons, pictures, icons) should have a hover effect and cursor pointer. User must intuitively understand that he can interact with them

2. Social icons in footer are nice and works as expected but I can suggest to wrap it in `<div />` tag or something else to a little bit increase "clickable" area.

3. Looks like I can send form with empty fields now (add validation to email also):
> The form shouldn’t submit empty

PR #838, opened 2023-12-21
task done

Reaffith:
[DEMO LINK](https://Reaffith.github.io/layout_dia/).


Reaffith:
horizontal scroll appears only in responsive mode in browser, if change size of browser window by hand it works good, dint know what kind of problem this it

nazarmatsevych (changes requested):
Good work

1. Page should not reload on submit, also clear form fields on button click
image

2. Remove [resize ](https://developer.mozilla.org/en-US/docs/Web/CSS/resize)
image

3. Add some hover effects to all interactive elements
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Need to fix the indent between these elements as on the design
image

2. No need to hardcode the width for these elements, the width should be responsive
image

3. The width for the burger menu should be 100%, also need to remove the scroll when this menu is open
image

4. Need to fix the left and right indents for this section as on the design
image

VitaliyBondarenko1982 (approved):
Good job.
Approve this time but left a few comments that suggest to fix during prepare your project for portfolio.
- on some width (on screen approxomately 1050px) not visible some part of footer by sides.
Screenshot from 2023-12-24 10-12-08
- check design and make inputs more higher (check distance between text or placeholder and line). For message field for user should be visible at leaset 2-3 rows.
- For textarea instead overflow hidden better to make [custom scrollbar](https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp)
- [fix autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/) for inputs
Screenshot from 2023-12-24 10-12-08

PR #849, opened 2024-01-03
Develop

pzwolenik:
[DEMO LINK](https://pzwolenik.github.io/layout_dia/).


anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-01-03 at 15 07 15
Consider fixing validation errors 
Screenshot 2024-01-03 at 15 04 21
Screenshot 2024-01-03 at 15 05 36
The layout seems broken for some devices. Also, consider opening an aside on the menu click

pzwolenik:
Hi!
For now, I changed as recommended the address element. 

I know it's the slider that's causing the error related to that white box on the right, but I've been thinking about it for a while, and I can't figure out where the error is. I suspect the JS code, but could I ask for help? Thanks :) 

Once I solve this problem, I'll happily add an additional menu - probably similar to layout miami?

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to fix it
image

2. Need to implement the burger menu when user click here
image

pzwolenik:
> Hi!
> 
> For now, I changed as recommended the address element. 
> 
> 
> 
> I know it's the slider that's causing the error related to that white box on the right, but I've been thinking about it for a while, and I can't figure out where the error is. I suspect the JS code, but could I ask for help? Thanks :) 
> 
> 
> 
> Once I solve this problem, I'll happily add an additional menu - probably similar to layout miami?

Hi! Sorry, can you first replay me to this message? Thanks ;)

volodymyr-soltys97 (changes requested):
A problem with a white line arises through this block because here the columns have a fixed width and they stretch our page, this must be fixed through styles
image

As for the menu then yes you can make it look like in miami

PR #855, opened 2024-01-09
Develop

vitalii-fedusov:
[DEMO LINK](https://vitalii-fedusov.github.io/layout_dia/)


DarkMistyRoom (changes requested):
Good start. Keep going and request the next review when the whole page is ready

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.
- content of this section should be centered horizontally
  
Screenshot 2024-01-14 at 18 43 32
- Suggest to [fix autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
Screenshot 2024-01-14 at 18 43 58
- make hover effect for social icons as in design
Screenshot 2024-01-14 at 18 44 37

DarkMistyRoom (approved):
Good job! Only two suggestions:
- would be good to add some hover effect to these cards
image

- these links should lead to real socials
image

PR #858, opened 2024-01-13
add page

manch0ffline:
[DEMO LINK](https://manch0ffline.github.io/layout_dia/)


SanyaBratashchuk (changes requested):
1.it would be better to implement mobile menu 
image

manch0ffline:
Але меню немає в макеті по дефолту. Його не потрібно робити

l4st1m0za (changes requested):
Great job done here! 

The only thing to fix is autocomlite styles in the email text input.

Also, as you mentioned before, menue is not default, but you can add one to show this things(photo) in there. You can make it as in MyBike landing!
image

PR #859, opened 2024-01-13
Develop

manch0ffline:
It is not necessary to create a menu for phones, since it is not specified in the task itself, and is not in the layout.

[DEMO LINK](https://manch0ffline.github.io/layout_dia/)


VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.
-  On form submit the fields should be cleared and no need any scrooll to top of page. Use: `<form onsubmit="this.reset(); return false;">`
-  in design no exists mobile menu, but suggest to implement it. it is will be just list of links (with similar functionality as menu for other landings).

SanyaBratashchuk (approved):

PR #863, opened 2024-01-23
solution

DanielYurchenco:
[DEMO LINK](https://DanielYurchenco.github.io/layout_dia/).

I'm sorry, but in some moments I did it as I saw, because the layout was not done very well:
there were no footnotes
there was no way aside should look like,
some sizes are not proportional to each other.
Thanks for understanding


Viktor-Kost (changes requested):
Nice!

To improve:
1. add a favicon
2. logo is a link
3. Make both 'hover' and 'unhover' be smooth. 
To reach this effect you should apply `transition` not to `:hover` but to static selector
Wrong Example: 
```
.button {
  color: #ff000;
  
  &:hover {
    color: #00ff00;
    transition: color 250ms;
  }
}
```

Correct Example:
```
.button {
  color: #ff000;
  transition: color 250ms;
      
  &:hover {
    color: #00ff00;
  }
}
```

4. form fields should be required
5. remove default input autocomplete styles
Screenshot 2024-01-23 at 16 45 49

6. set a correct input type for email
7. `apply` button leads to 404 page
8. check font styles
Screenshot 2024-01-23 at 16 48 50

9. inputs and button should have different width, check mockup
Screenshot 2024-01-23 at 16 53 03


Recommendations:
1. content should be centered, responsive and fill the full parent space
Screenshot 2024-01-23 at 16 49 53

2. Suggest making sure that on desktops, the content section doesn't spread across the entire screen. Instead, it should have a maximum width. If there's a background image, image shouldn't be limited (only a section content). To do this, you need to apply the right styles in your CSS.

**Why it's important:**

_**Fits well:**_ Setting a maximum width ensures that your design looks good on desktops without the content becoming too wide.

_**Easy to read:**_ Keeping the content width in check makes it easier for people to read, as really long lines of text can be hard to understand.

_**Same look:**_ Having a maximum content width helps maintain a consistent design, especially when using background images that go beyond the content area.

_**Easy to organize:**_ Limiting the content width makes it simpler to organize your content on the page. This makes it easier for developers and designers to manage.

maxim2310 (changes requested):
Yor demo link is not working, please fix it and make request review again

Viktor-Kost (changes requested):
### GJ!

### Previous comment #8
Check font styles (it's better to check all over the page)
Yours:
Screenshot 2024-01-24 at 16 40 26

Figma:
Screenshot 2024-01-24 at 16 41 24

### To improve:
1. I see you've improved a form logic. Yet page shouldn't reload after submit

IvanFesenko (changes requested):
- Form validation do not works
https://github.com/mate-academy/layout_dia/assets/61904995/6ec3f2a0-a3fd-4027-8801-e1272d7625ca

etojeDenys (approved):
it's should be the textarea
image


looks good

PR #868, opened 2024-01-25
Develop

ilyaivasyk:
[DEMO LINK](https://ilyaivasyk.github.io/layout_dia/).


denys-danyliuk (changes requested):
double check your demo link and your landing overall with a design because now it doesn't match it
also make sure to go through checklist before requesting a review

PR #871, opened 2024-01-26
add task solution

OksanaZderko:
[DEMO LINK](https://OksanaZderko.github.io/layout_dia/)


l4st1m0za (approved):
Really awesome job here! The only thing you can improve is to make so that insta, twiter and facebook icons would really link to some page on that services, but in all, great job!

PR #875, opened 2024-01-30
solution

RVDotsenko:
[DEMO LINK](https://RVDotsenko.github.io/layout_dia/)


etojeDenys (changes requested):
1. looks strange here
image

2. looks like you're using the incorrect font 
image
image

TarasHoliuk:
1. There is still a problem with fonts (maybe you forgot to re-deploy your solution):
image

image

image

image

2. You did an amazing thing here! But the images from the next section are displaying over the content you've added. It would be nice to fix it
image

3. After clicking `Learn more` in the header:
image

Maybe add some scroll inside this section that appears, etc. Now the content is just cut down. But anyway - great job :+1:

TarasHoliuk (changes requested):
Very good job :+1: 

Make sure you re-deployed your solution after changes, also added a few additional comments

DarkMistyRoom (changes requested):
Great job! But there are still some comments:
- would be great to add some hover effect to images
- please check the font it looks different in the mockup  
image

- would be great to make this background white as above
image

etojeDenys (approved):
amazing

PR #876, opened 2024-01-30
Develop

alionazherdetska:
[DEMO LINK](https://alionazherdetska.github.io/layout_dia/).


etojeDenys (changes requested):
1. disable the scroll when menu is open
image

2. add the favicon 
image


3. try to remove the default autocomplete styles here
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Add the actual links for these social networks
image

2. Why did you remove the autocomplete for the inputs?
It does not need to be excluded, but only needs to be additionally styled
image

3. Add a transition when this menu appears on the page
image

volodymyr-soltys97 (changes requested):
To improve:
1. These links should be open in the new tab
image

volodymyr-soltys97 (approved):
Nice work 🔥

PR #877, opened 2024-01-31
solution

KachVl:
[DEMO LINK](https://kachvl.github.io/layout_dia/)


vadiimvooo (changes requested):
Good job! A few changes required:

1. It would be good to have some hover effect with this images:
image
2. You should add the transition effect on transform property to make the transform process smothly.
image
3. Also here is wrong text align, please check the right one on mockup:
image

etojeDenys (changes requested):
1. stretch slider on mobile
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

need to remove autofill style 
image

DarkMistyRoom (changes requested):
Good! You're almost there

- don't need to lead to the top on the form submitting, only clear the form
- all socials should open a new tab
image

KachVl:
@DarkMistyRoom 
Checklist, paragraph 9: When you try to send the form there is no 405 error and the form is automatically cleared after submit **and is scrolled to the top of the page** or the page is reloaded

PR #879, opened 2024-02-02
task solution

h1deeeee:
[DEMO LINK](https://h1deeeee.github.io/layout_dia/)


sTorba24 (changes requested):
Nicely done! Lets improve it :)

1. Add some hover effect to navigation links

2. use cursor-pointer
image


3. add max-height here, no need to stretch on huge screen, it looks weird
image


4. remove autocomplete styles (blue background)
image


5. prevent default behavior on form submit and reset it after

6. add hover effects to links
image


7. here too 
image


8. button should have the same width as inputs on mobile
image


now: 
image

SanyaBratashchuk (changes requested):
1. horizontal scroll on mobile
image

PR #881, opened 2024-02-04
Implemented AIR landing

Kovachhh:
[DEMO](https://kovachhh.github.io/layout_dia)

[DESIGN](https://www.figma.com/file/7qwsWggv9BAxMi2VPhBuPr/Air-(formerly-Dia)?node-id=9138%3A35)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The burger menu should stretch to the entire width and height of the browser window, and you should also remove the scroll when it is open
image

2. Need to change the autocomplete styles
image

3. You also need to correct the attributes for the form and inputs as written here on the screen
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, autocomplete doesn't work
image
image

Need to remove the scroll of the page, when this menu is open
image

Kovachhh:
I have fixed attributes of all fields based on previous comment. Okay, I return 'autocomplete' attribute for form

IvanFesenko (changes requested):
Looks good, check this moments:
- All fieds in form should be required
- Some thing weird in navigation
https://github.com/mate-academy/layout_dia/assets/61904995/50eb2826-cb36-408c-91ae-40e09283fce4

anastasiiavorobiova (approved):
Great work!

PR #888, opened 2024-02-16
add task solution

vneholiuk:
[DEMO LINK](https://vneholiuk.github.io/layout_dia/).


sTorba24 (changes requested):
Good work! Lets improve it

1. huge x-scroll here, also image is cropped
image

2. center links vertically
image


3. hover effect visible only on twitter(x), should be for all
image


4. white line at the bottom of page
image

SanyaBratashchuk (approved):

PR #891, opened 2024-02-19
add task solution

lmuias:
[DEMO LINK](https://lmuias.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to change the autocomplete styles for the input fields
image

2. Make all these fields are required before send and after click the send you need to clear the form fields
image

3. Add actual links for these social networks and fix the facebook icon as on the design
image

4. The logo should be link on the footer
image

5. This menu should be width 100%, also, you need to remove the scroll when this menu is open
image

TarasHoliuk (approved):
You did an amazing job - approved :heavy_check_mark: 

One suggestion (you may do it before adding the landing page to your portfolio): make blue buttons full-width for mobile version (it will look better):
image

lmuias:
done)

PR #896, opened 2024-02-22
Implement air landing

deezhizyu:
[DEMO LINK](https://deezhizyu.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to change the color for twitter icon and should to open these links in the new tab
image

2. Need to change the autocomplete styles for the input fields. Also, after clicking send, the page should not scroll up
image

3. This link should be open in the new tab
image

4. Need to fix it on the mobile and tablet versions
image

5. Check it on the design
image

6. Need to implement the burger menu when user click here
image

Viktor-Kost (changes requested):
GJ!
Let's check task requirements:
1. The "Learn more" button should lead to the closest block (Our expertise)
2. A landing page is implemented strictly according to the design in Figma
Check font styles for this button 
Screenshot 2024-02-23 at 13 39 42

3. In the Vision, Passion, Results section, the “Apply” button also should lead to the contact form

VitaliyBondarenko1982 (approved):
Good job. Approve.
Found one issue which you can fix during prepare project for portfolio.
- The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
`<form onsubmit="this.reset(); return false;">`

PR #899, opened 2024-02-26
Develop

MatheusStefan:
[DEMO LINK](https://MatheusStefan.github.io/layout_dia/).


src/styles/blocks/strategic-agency.scss
```diff
@@ -0,0 +1,326 @@
...
+    }
+  }
+  
+  .carousel {
+    position: relative;
+    width: 100vw;
+  
+    @include on-small-desktop {
+        width: 502px;
+        height: 680px;
```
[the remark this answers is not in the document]
MatheusStefan: In that case, inside the carousel, I tried to not hardcode every pixel on the classes, but in general it was not working. So the only option left was hardcode it, unfortunately :(

src/index.html
```diff
@@ -1,19 +1,384 @@
...
+
+    <section class="strategic-agency" id="about-us">
+      <div class="strategic-agency--top">
+        <h1 class="strategic-agency--top-title">Strategic Agency</h1>
+        <span class="strategic-agency--top-content">We believe in the power of bold ideas that can solve business challenges.</span>
+        <button class="btn" href="#">Learn more</button>
+      </div>
+
+      <div class="strategic-agency--bottom">
+        <div class="carousel">
```
EdPirro, line 61: This carousel does not seem to be working properly on 1440p screens [Reference](https://github.com/mate-academy/layout_dia#:~:text=Large%20screens%202560px)

image

src/index.html
EdPirro, line 1: Try to center your content so that the page doesn't look broken on screens in between breakpoints

image

src/styles/blocks/cta.scss
```diff
@@ -0,0 +1,91 @@
...
+      @include on-small-desktop {
+        padding: 0 200px;
+      }
+
+      @include on-desktop {
+        padding: 0 320px;
+      }
+    }
+
+    .btn {
```
EdPirro, line 71: Avoid using relative position between elements to apply styles; this leads to highly coupled code; which is harder to maintain and less reusable

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,44 @@
...
+  border: none;
+  border-radius: 8px;
+  width: 280px;
+  height: 50px;
+  margin-bottom: 48px;
+  padding: 0;
+
+  cursor: pointer;
+
+  &:hover {
```
EdPirro, line 39: Hover transitions should be smooth both entering and leaving hover; this issues doesn't happen just here; please double check every hover element for usability

EdPirro (changes requested):
Great effort, @MatheusStefan! Just a few more touch-ups and your code should be ready to ship 🚀

src/index.html
MatheusStefan: I'm aware of that, you just have to reload the page and it should work fine.. I used JS on this project.

src/scripts/main.js
```diff
@@ -1 +1,37 @@
...
+const moveToSlide = (track, currentSlide, targetSlide) => {
+  track.style.transform = 'translateX(-' + targetSlide.style.left + ')';
+  currentSlide.classList.remove('current-slide');
+  targetSlide.classList.add('current-slide');
+};
+
+prevButton.addEventListener('click', (e) => {
+  const currentSlide = track.querySelector('.current-slide');
+  const prevSlide = currentSlide.previousElementSibling;
+
```
pedro-ruas, line 28: This gives a console error whenever you scroll to the end of your slides and try to continue after it.
You could add a test here to, in case the slide you're trying to move to does not exist, move it to the last one (like an infinite slider"

src/scripts/main.js
```diff
@@ -1 +1,37 @@
...
+  moveToSlide(track, currentSlide, prevSlide);
+});
+
+nextButton.addEventListener('click', (e) => {
+  const currentSlide = track.querySelector('.current-slide');
+  const nextSlide = currentSlide.nextElementSibling;
+
+  moveToSlide(track, currentSlide, nextSlide);
```
pedro-ruas, line 36: Same error happens here.

src/styles/blocks/cta.scss
```diff
@@ -0,0 +1,91 @@
...
+      width: 280px;
+      height: 50px;
+      text-decoration: none;
+      color: #fff;
+      font-size: 15px;
+  
+      &:hover {
+        color: #2060F6;
+  
+        transition: all 0.3s ease-out;
```
pedro-ruas, line 82: You added this *transition* property only for the hover state, this creates a smooth in but hard out transition to your element.
That happens because the moment the user stops hovering the element, it loses it's *hover state* and fallsback to the normal state, where there isn't a defined *transition* property. So your element looks smooth while the hover starts, but not when it ends.

The fix is simple, just move this **transition** property to **the element's normal state, outside the &:hover**.

And I have checked your Demo Link and this is happening with all your transitions, better check the rest too

src/scripts/main.js
```diff
@@ -1 +1,37 @@
...
+
+slides.forEach(setSlidePosition);
+
+window.addEventListener('resize', (e) => {
+  slides.forEach(setSlidePosition);
+});
+
+// eslint-disable-next-line no-shadow
+const moveToSlide = (track, currentSlide, targetSlide) => {
+  track.style.transform = 'translateX(-' + targetSlide.style.left + ')';
```
pedro-ruas, line 20: You could change this notation to:
```
track.style.transform = `translateX(-${targetSlide.style.left})`
```
In this way, instead of concatenating strings, you would be adding a variable to a string.

src/scripts/main.js
```diff
@@ -1 +1,37 @@
-'use strict';
+const track = document.querySelector('.carousel__track');
+const slides = Array.from(track.children);
+const nextButton = document.querySelector('.carousel__button-right');
+const prevButton = document.querySelector('.carousel__button-left');
+
+const slideWidth = slides[0].getBoundingClientRect().width;
```
pedro-ruas, line 6: if I'm not mistaken, the only place where you're using this *slideWidth* variable is inside *setSlidePosition* function.
If this is the case, move this variable declaration inside the function to prevent some errors.
Like what is happening in this image here:
image

pedro-ruas (changes requested):
Some improvements to be done with transitions and JS

EdPirro (changes requested):
Great effort, @MatheusStefan! Your landing page looks good to go! You fixed all the requested changes, but you the deployed site seems to be outdated!

Please, make sure to deploy the latest version of your code for review, if you're facing issues, just let me know!

EdPirro (approved):
Awesome job, @MatheusStefan! Keep it up! 🚀

PR #902, opened 2024-03-01
solution for Landing page Dia

pindel-kateryna:
[DEMO LINK](https://pindel-kateryna.github.io/layout_dia/)


sTorba24 (changes requested):
Good work! Try to redeploy  your work, now it leads to 404 page :(

vadiimvooo (changes requested):
Good job. A few things needs to be fixed:
1. This link should have transition as well
image
2. It would be better to use hover effect when you hover all card instead of making this effect on hover the element.
image
3. Footer links on tablet should be fixed according to the mockup.
image

TarasHoliuk (approved):
Nicely done! 2 themes implemented :fire: 

I'd rather make the theme change button the icon: image
[Icons](https://mui.com/material-ui/material-icons/?query=da), and move it to the header. But up to you :slightly_smiling_face:

PR #908, opened 2024-03-10
Develop

hardixxxxx:
[DEMO LINK](https://hardixxxxx.github.io/layout_dia/)
have problem with npm run deploy


volodymyr-soltys97 (changes requested):
Demo link, if you have any error with `npm run deploy` feel free in the fe_chat
image

hardixxxxx:
i think i have problem with slider position, bcoz i dont know how i can calculate width, when i add max width 1600px and margin inline auto, for fix this problem add to header overflow hidden.
and my theme switcher not working in github pages idk why, in my live server this worked

SanyaBratashchuk (changes requested):
1. Remove this link. It would be better to replace it with smth like switcher 
image
2. Fix autocomplete styles 
image

etojeDenys (approved):
lgtm

PR #909, opened 2024-03-14
add solution

Opokhvalenko:
https://Opokhvalenko.github.io/layout_dia/


anastasiia-tilikina (changes requested):
You did great job 🔥  Well done with slider!
Here are some issues we need to fix

1. Check font-family everywhere in your page. It should be either Poppins of Open sans, but somewhere it's Helvetica of even Times

image

image

Check it everywhere

2. Add smooth scrolling to the page:
```scss
html {
  scroll-behaviour: smooth;
} 
``` 

3. add hover effects for all interactive elements
4. here only one icon changes color on hover
image

4. Link contact us doesn't work
5. Icons on mobile menu are broken
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

image

3. Add actual links for these social networks
image

4. Remove it 
image

5. You need to fix font weight for the text in button, check it on the design
image

6. Fix the indent between the button and slider
image

sTorba24 (changes requested):
Good work! 

1. add hover effects on buttons
image

2. add target _blank attribute for these links
image

VitaliyBondarenko1982 (approved):
Good job.
Mentioned a couple points that you can fix during prepare project for portfolio:
- add transtion on elements with hover effect. at lease buttons and social links
chrome-capture-2024-3-16 (2)
- disable scroll on page when menu is open.

PR #911, opened 2024-03-19
Develop

ArtemZakhar:
I have doubts about the slider correctness implementation.
I understood incorrectly the task.

I would like to make the theme switcher also.
But firstly would be great to fix all mistakes.

[DEMO LINK](https://ArtemZakhar.github.io/layout_dia/)


src/scripts/components/menu.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+export default function Menu() {
+  const body = document.querySelector('.page');
+  const menu = document.querySelector('.navigation');
+  const navList = document.querySelector('.navigation__wrapper');
+  const links = document.querySelectorAll('.navigation__nav-link');
+  const closeBtn = document.querySelector('.navigation__close');
```
DanilWeda, line 8: Better variant use `js-` selectors for using in JS logic. 
For example `.js-page .js-navigation`

src/scripts/components/menu.js
```diff
@@ -0,0 +1,28 @@
...
+
+  menu.addEventListener('click', () => {
+    body.style.overflow = 'hidden';
+    navList.classList.add('navigation__wrapper--active');
```
DanilWeda, line 12: I think you can add active ( not as a modifier )

src/scripts/components/form.js
```diff
@@ -0,0 +1,16 @@
...
+    const formData = new FormData(form);
+
+    for (let [name, value] of formData.entries()) {
+      console.log(name, ':', value);
+    }
+
+    form.reset();
+    window.scrollTo(0, 0);
+  };
+}
```
DanilWeda, line 16: You can not export form, and write all in a global scope (as alternative)
```suggestion

  const form = document.querySelector('.contacts_form');

  form.onsubmit = (event) => {
    event.preventDefault();

    const formData = new FormData(form);

    for (let [name, value] of formData.entries()) {
      console.log(name, ':', value);
    }

    form.reset();
    window.scrollTo(0, 0);
  };
```

src/scripts/components/menu.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+export default function Menu() {
```
DanilWeda, line 3: Also for menu

src/scripts/main.js
```diff
@@ -1 +1,7 @@
-'use strict';
+import Menu from './components/menu';
+import Slider from './components/slider';
+import Form from './components/form';
```
DanilWeda, line 3:
```suggestion
import './components/menu';
import './components/form';
import Slider from './components/slider';
```
Then, if you change export logic you can simple import file inside main)

src/styles/blocks/title.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+.title--s {
+  text-align: center;
+  font-size: 32px;
+  font-weight: 600;
+  line-height: 150%;
+  color: $grey-900;
+}
```
DanilWeda, line 19: `--s` modifier without element

src/styles/blocks/title.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.title--l {
```
DanilWeda, line 21: Also in this place

src/scripts/components/slider.js
DanilWeda, line 1: I think, for start, you can simplify your custom slider.
Please make a slider that has buttons prev/next, which are controls.
Make auto speed for the next slide. If it is the last slide, go to the first, and in the reverse, you need the same logic. Also, controls inherit this logic. Then we can update your slider and add new opportunities. 
Many thanks)

DanilWeda (changes requested):
Good job. 
Check some comments for improvements! 

Also layout comments:
1. Add hover for actions elements 
image


2. Fix size of link element 
image

3. Also add hover for menu and buttons 
image


Many thanks)

src/scripts/components/slider.js
ArtemZakhar: Thank you for your comments) 
Some of them are confusing. For example about js-selectors, I have never faced such approach)

And I didn't get about the slider. 

I thought that we had to implement the opportunity to change slide by scroll. 
That's why the slider has strange behavior on the tablet and phone size.
It was a challenge for me, so I want to fix recent slider and add touchmove event instead of mousemove on smaller screens. Never try to use it.

`Make auto speed for the next slide. If it is the last slide, go to the first, and in the reverse, you need the same logic. Also, controls inherit this logic. Then we can update your slider and add new opportunities.`

this one has been implemented already. I think, maybe while you changing the size of a screen - slider displays wrong. That is because I calculate the width of the slider according to the screen width. You just need to reload the page.
Or I understand incorrectly your issue. 
Please explain.

ArtemZakhar:
> 2. Fix size of link element
>    image

> Many thanks)

@DanilWeda  I can't remember how to forbid flex to stretch on the main axe( Justify-content and items don't work(
Only because I read yesterday about inline-flex - I used this strange composition to fix your second issue(

src/scripts/components/menu.js
ArtemZakhar: Oleksandr Ihnatov helped me to understand the issue. 
At first, I was confused, but now understand that all your replies are awesome!)

src/scripts/components/slider.js
ArtemZakhar: @DanilWeda  what will be better solution? 

I fixed bug with overflow on mobile devices and added swipe opportunity.

I can make it simpler by using opacity toggler for slider children or leave my attempt.

src/styles/blocks/theme-switcher.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.theme__switcher::after {
+  content: '';
+  display: block;
+  width: 50%;
+  padding-bottom: 50%;
+  border-radius: 50%;
+  background-color: var(--backgroung-color);
+}
```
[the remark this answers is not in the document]
ArtemZakhar: @alexander-ignatow this one I did not get(
please explain.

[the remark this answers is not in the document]
ArtemZakhar: sorry, i get it)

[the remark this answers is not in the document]
ArtemZakhar: Thank you! I have found several more)

alexander-ignatow (approved):
Let's proceed with it 👍

PR #914, opened 2024-03-24
solution

emberlast1:
[DEMO LINK](https://emberlast1.github.io/layout_dia/)


GUSILLUS (changes requested):
GJ! But:

1. Try to change hover effect for nav link
image

2.  Same
image

3. Fill all button by blue color
image

4. Fix hover here, it look like broke UI. Just use scale or change color of icon
image

5. Add padding 
image

6. Try to place tablet sidebar label in red square
image

etojeDenys (approved):
good job, but better to disable the scroll when menu is open

PR #920, opened 2024-04-04
Develop

YuliiaCat:


volodymyr-soltys97 (changes requested):
Hi, there is no demo link here, also you need the demo link to the PR desctription
image

l4st1m0za (changes requested):
Great job! The only things to fix are:
1) Remove autofill styles in the email input. Here is a link which will help you https://css-tricks.com/almanac/selectors/a/autofill/
image

2) Remove scroll from the menu

loralevitska (changes requested):
Almost perfect) Let's improve for more practice

- add hover effect to links in menu
image
- for mobile mode stretch slider to full width
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. Add a favicon on the page
image

2. Fix slider position on the tablet version, check it on the design
image

3. The left and right indents must be are same here
image

4. Remove the fixed width in px for the slider and use `width: 50%` instead, it will be better
image

volodymyr-soltys97 (changes requested):
Almost done!
Still not fixed from the previous review
image

etojeDenys (approved):
GJ

PR #926, opened 2024-04-26
add solution

Mariiaantoniv:
[DEMO LINK](https://Mariiaantoniv.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. These cards must be the same width
image

2. Need to change the autocomplete styles
image

3. Add some hover effects for these clickable elements
image

4. Remove padding for this block on the mobile version
image

nazarmatsevych (approved):
Well done 👍

PR #929, opened 2024-05-01
-solution

RuslanSkl:
[DEMO LINK](https://RuslanSkl.github.io/layout_dia/)


GUSILLUS (changes requested):
GJ! But:

1. Try to fix this
image

2. Make slider workable
image

3. Consider change the hover logic here. Try to change only  image size
image

4, In need to be a social link not a github link
image

PR #932, opened 2024-05-05
Dia

Beliyandr:
https://beliyandr.github.io/layout_dia/


etojeDenys (changes requested):
1. add some favicon here
image

2. should be links
image

3. should be rounded
image
image

4. 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve: 
1. Add actual links for these social networks 
image

2. You need to remove the horizontal scroll on the mobile version 
image

Esceype (changes requested):
GJ!
Need to fix:
- Main logo in header redirect us to not found page:
image;
- remove default background color after autofill:
image
- submit button inside contactUs form should clean all inputs without reload the page;
- scroll should be disabled when menu is open:
image
- all buttons and icons/images should have hover effects and cursor pointer;
- page on device smaller then 320px width should looks the same as on mobile on 320.
image

etojeDenys (approved):
lgtm

PR #938, opened 2024-05-21
add task solution

VladKugot:
[DEMO LINK](https://VladKugot.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page 
1. The cards should decrease as smoothly as they increase
image

2. Add cursor pointer for this button 
image

3. After submit form don't need to refresh page and scroll to top, you need to clear the form fields only 
image

anastasiia-tilikina (approved):
Well done!

PR #944, opened 2024-06-02
Develop

hannasavka:
- [DEMO LINK](https://hannasavka.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Hi, there are no styles, you need to make a deploy again and check it after on the demo link 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page 
image

3. Add cursor pointer for all buttons

4. You need to open additional links in the new tab 
image

5. Add transition when this menu appears on the page and add indent between the logo and menu items 
image

6. The menu should be vertical like on the mobile version
image

volodymyr-soltys97 (changes requested):
Almost done!
1. This comment still not fixed from the previous review 
image
image

2. Hover styles should also disappear as smoothly as they appear, now it happens immediately, it needs to be fixed everywhere
image

3. Add hover effects for these elements 
image
image

vadiimvooo (approved):
Well done

PR #947, opened 2024-06-18
add project dia

gorbachovam00:
[DEMO LINK](https://gorbachovam00.github.io/layout_dia/).


vadiimvooo (changes requested):
Good job, there are a few changes to fix:
1. This section have to be for the full width
image
2. This elements has to be in the form of a round.. Please fix it.
image
image
3. Footer section should fill full width as well
image
4. For this link you have to add `target="_blank"` attribute
image
5. These elements should be links which follows the appropriate social media.
image

anastasiiavorobiova (approved):
Great work!

PR #948, opened 2024-06-23
add task solution

AndriiZakharenko:
-  [DEMO LINK](https://AndriiZakharenko.github.io/layout_dia/)


DanilWeda (approved):
Good job. 
Check some UI moments:

1. Icon in head tag 
image

2. Also check font-weight which your use. Maybe I miss something, but I think their have differences. 

3. You can compress photo for better first load.

Final, I checked your code - awesome job! 
Thanks you for request. I approved with a small comments. 
Have a nice day!

PR #949, opened 2024-06-25
Solution

Galers:
[DEMO LINK](https://Galers.github.io/layout_dia/)


etojeDenys (changes requested):
1. why do you need this margin on mobile
image

2. better to disable the scroll if the menu is open

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to change the autofill styles for the form fields 
image

2. Don't need to scroll page to top after submitting form, you need only clear form fields

3. Add transition for hover effects here 
image

4. These links should lead to the relevant sections
image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #954, opened 2024-07-05
add task solution

MK030191:


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements on the page 
image
image

2. The slider should be pressed against the right side of the browser window 
image

3. Remove the scroll when the burger menu is open 
image

4. Add scroll behavior smooth to the html selector

PR #955, opened 2024-07-08
bug fixes

MK030191:


etojeDenys (changes requested):
1. 
image


2. disable the scroll when menu is open

PR #957, opened 2024-07-10
add task solution

MK030191:


anastasiia-tilikina (changes requested):
Well done!
Things we need to fix:

1. you need to fix bg image for this section on large screen
image

2. Add hover effects on these links
image

3. gap here is too big
figma:

image

your page:
image

4. disable page scroll when menu is open

anastasiia-tilikina (changes requested):
.

PR #964, opened 2024-07-22
add solution

renata-aline:
[DEMO LINK](https://renata-aline.github.io/layout_dia/)


src/scripts/main.js
```diff
@@ -1 +1,61 @@
 'use strict';
```
gitlherme, line 1: This whole file is commented. If you are not using the code, please delete it.

gitlherme (changes requested):
Hello, Renata! Very good job with your code. I have some points about the behavior of the page. Some items in README are missing on your project. Can you please check and adjust it?

- Add a smooth scroll for the whole page; (When I click in some link, doesn't have smooth scroll)
- When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded. (Your page is not scrolling after submit).
- OPTIONAL: The slider is not working, if you want, make it happen! 😄 

Thank you!

src/scripts/main.js
```diff
@@ -1 +1,44 @@
...
+
+  window.scrollTo({
+    top:0,
+    behavior:'smooth'
+  })
+})
+
+
+
+  document.addEventListener('DOMContentLoaded', function () {
```
pedro-ruas, line 15: I understand your idea here, but I will give you a suggestion that could improve on visual:
Because you are using a class "active" to select what image is showing, when using prev and next, the image "jumps" changing without an animation. that isn´t a real problem, but everything else on the website uses animation, and this this looks sorta out of place.
Instead of using the active class, you could use currentIndex to `transform: translateX(${currentIndex * 100%)` in the slider. To have a sliding animation when changing images.

That, of course, is not mandatory, just an idea if you want to implement it

pedro-ruas (changes requested):
Great work! I see you have corrected everything Guilherme requested, but I see an issue when changing the first slider, one image is bigger than the others and makes the slider change depending on the current image.

Besides that, I left just a suggestion in your code.

pedro-ruas (approved):
Great Job!

PR #967, opened 2024-07-27
add task solution

svitlanak31:
- [DEMO LINK](https://svitlanak31.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements
image

2. Add cursor pointer for this button 
image

3. Where is the facebook icon? 
image

4. The burger menu should be the entire width and height of the screen, you should also remove the scroll when it is open
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. This comment still not fixed. Add cursor pointer to button in the form 
image

2. Change the autofill styles for the form fields
image

3. Fix background for the logo on the burger menu 
image

4. Add transition when the burger menu appears on the page 
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #974, opened 2024-08-20
Ready header for mobiles & tablets

sborichevskyi:
[DEMO LINK](https://sborichevskyi.github.io/layout_dia/).
  i've made the header for mobiles and tablets. I'd be glad to hear your recommendations. BTW, when i came across header on desktops, i understood, that header content (except top bar) should be displayed with grid, right?


volodymyr-soltys97 (changes requested):
Good start!
To improve:
1. Add a favicon on the page
image

2. Need to implement the burger menu when user is clicking on this icon 
image

3. Keep going with the next sections

PR #987, opened 2024-09-15
Develop

jooaasss:
[DEMO LINK](https://jooaasss.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Change the autofill styles for the form fields
image

image

3. Add the horizontal padding for the content 
image

anastasiiavorobiova (changes requested):
Great work! Check the layout for 1024px. It should match the Figma

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. The gap between these blocks must be are same 
image

2. Don't need to refresh page after submitting form and scroll to top, you need only reset form fields 
image

3. Add hover effects here 
image

4. Don't need to write fixed width for these blocks on the mobile version 
image

etojeDenys (changes requested):
amazing idea with change theme!

1. it should be the textarea tag
image

2. the button should have type submit
image

3. and then update your form to prevent reloading
`<form onsubmit="this.reset(); return false;">`

etojeDenys (approved):
🤩

PR #988, opened 2024-09-15
Develop

tuhusova:
[DEMO LINK](https://tuhusova.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your page
1. Add hover effects and cursor pointer for all clickable elements on the page 
image
image

2. Change the autofill styles for the form fields
image

3. The user cannot be able to change size of the textarea field 
image

4. These elements must be links and add hover effects here
image

5. Add transition when this menu appears on the page 
image

SanyaBratashchuk (approved):
1. apply transitions for hover effects

PR #997, opened 2024-10-12
add task solution

iirk1:
[DEMO LINK](https://iirk1.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page 
image

2. Fix width for these blocks 
image

3. The width must be are same everywhere and add hover effects here
image

4. Add hover effects for all clickable elements 
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. Make all these fields are required before sending 
image

2. Open these additional links in the new tab 
image

3. These links must be lead to the relevant sections and add scroll behavior smooth on the page 
image
image

4. Remove max width for these blocks, use 100% instead 
image

volodymyr-soltys97 (changes requested):
Almost done!
You need to implement the menu when user clicks here 
image

anastasiiavorobiova (approved):
Awesome!

PR #999, opened 2024-10-19
make header

Ga1dar:
[DEMO LINK](https://Ga1dar.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Hi, the header looks like strange, if you need some help feel free in the fe_chat 
image

Fix gap between these elements 
image

The menu should span the entire page in width and height
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
- suggest to [fix autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/) and [make custom scrollbar](https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp).
- prevent resize textarea
Screenshot 2025-03-01 at 19 19 13

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add hover effects for all clickable elements on the page. 
image

2. Why did you implement scroll here? Check it on the design 
image

3. Need to open these additional links in the new tab 
image

etojeDenys (approved):
good job overall

PR #1004, opened 2024-11-01
add task solution

hrebinets:
[DEMO LINK](https://hrebinets.github.io/layout_dia/).


etojeDenys (changes requested):
1. should have the same indents
image

2. try to add some favicon here
image

3. let's open the address link on a new tab
image

VitaliyBondarenko1982 (approved):
Looks good

PR #1005, opened 2024-11-11
Develop

pdabizha:
- [DEMO LINK](https://pdabizha.github.io/layout_dia/)
- [MOCKUP](https://www.figma.com/design/7qwsWggv9BAxMi2VPhBuPr/Air-(formerly-Dia))?node-id=9456-23&node-type=frame&t=m2tBU4Nt3OQVcKqW-0)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Change the autofill styles for the form fields. Also, after submit the form you don't need refresh page, you need only reset form fields
image

image

3. You need to implement the burger menu when user clicks here 
image

2.

SanyaBratashchuk (changes requested):
burger menu is not implemented

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
The autofill styles must be are same as default styles 
image

volodymyr-soltys97 (changes requested):
Hi, the autocomplete styles should be the same as when we manually enter the data, that is, you need to change the background color and the text color
image

SanyaBratashchuk (approved):

PR #1009, opened 2024-11-25
dia landing page

leraMeleshchenko:
[DEMO LINK](https://leraMeleshchenko.github.io/layout_dia/)


SanyaBratashchuk (changes requested):
1. section looks wrong on wide screens
image
2. page shouldn't be reload after form submit
3. make all fields required
4. implement burger menu on mobile
5. center these cards 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.

- The page shouldn't be reloaded on form submit, but the fields should be cleared. 
Use: `<form onsubmit="this.reset(); return false;">`

- add hover effect (change color) for nav links in footer.

VitaliyBondarenko1982 (approved):
Well done

PR #1014, opened 2024-12-01
solution

1luki9901:
[DEMO LINK](https://1luki9901.github.io/layout_dia/)


danon321 (changes requested):
Demo link daje bład 404

Zibi95 (changes requested):
There is 404 on DEMO link and. I can't see any style files commited.

PR #1015, opened 2024-12-04
Develop

Bezushk0:
[DEMO LINK](https://bezushk0.github.io/layout_dia/)


etojeDenys (changes requested):
better to disable the scroll in this case, but looks good in general
image

etojeDenys (approved):
good job

PR #1019, opened 2024-12-10
add task solution

Uandrik:
[DEMO LINK](https://Uandrik.github.io/layout_dia/)


Uandrik:
Why there is an error 404 here?

volodymyr-soltys97 (changes requested):
Hi, you need to run command `npm run deploy` in terminal, if you need some help feel free in the fe_chat 
image

TarasHoliuk:
1. Fix layout in this section:

image
image


2. There should be some margin (check figma) between the title and subtitle:

image

3. Add hover effect to links and nav links (in the header as well):

image

4. Make blue buttons 100% width on the mobile version. No need to hardcode width - in the figma we see an example of mobile version design on 320px. But it should look the same on let's say 400px too (i.e. should be full-width):

image

Also, add a hover effect to such buttons.

5. Add placeholders:

image

6. I know there is no design for a mobile menu, so it's a suggestion. But it'll look better if you decrease the font size for links (make it the same as in the footer) add the logo at the top (same as in the header) and close "X" button in the top right angle

TarasHoliuk (changes requested):
GJ 👍 
Check comments

Uandrik:
https://snipboard.io/0WCP5a.jpg I don't see a problem with placeholders.

Sorry, now I see, but I don't know why.

VitaliyBondarenko1982 (approved):
Looks good.

PR #1021, opened 2024-12-18
Landing dia page created

vlavrentieva:
I created a page without JavaScript interactivity (without changing the color of the page theme and without a slider), as the mentors explained, this will need to be added later when I'll create a portfolio
https://vlavrentieva.github.io/layout_dia/


etojeDenys (changes requested):
please don't forget to check and confirm that your page looks the same as the mockup in figma
https://www.loom.com/share/10d4a049cea04b799e089f759609d074

vlavrentieva:
You mean blocks "Who we are", "TESTIMONIALS", "Vision, Passion, Results" on screens larger than 1024 pixels?

PR #1025, opened 2024-12-29
add task solution

joaoguilherme000:
[DEMO LINK](https://joaoguilherme000.github.io/layout_dia/).


joaorpereira (changes requested):
The following points in the checklist are missing:

1. The form shouldn’t submit empty
2. When you try to send the form there is no 405 error and the form is automatically cleared after submit and is **scrolled to the top of the page** or the page is reloaded
3. the hover animation should be applied when we hover in the **Learn More** text as well, and not just only when hovering the icon:
Screenshot 2025-01-03 at 10 11 54

4. Adjust margin on mobile:
Screenshot 2025-01-03 at 10 14 43
Screenshot 2025-01-03 at 10 15 10

REALLY GOOD WORK! JUST SMALL IMPROVEMENTS

PR #1029, opened 2025-01-07
Develop

diachkinainna:
[DEMO LINK](https://diachkinainna.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Fix width for the slider image, check it on the design 
image

2. Add transition for hover effects here 
image

3. Check positions here on the design 
image

4. Add cursor pointer here 
image

5. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. The gap between these elements is too big, check it on the design 
image
Design 
image

2. The color should be white here 
image

3. Add cursor pointer and some hover effects here 
image

4. The horizontal indents should be same everywhere 
image
image
image

Design 
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1037, opened 2025-01-28
Develop

MaksymMohyla:
https://MaksymMohyla.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Fix the background color for the menu as on the page, it's looks different now 
image
image

2. Also, you need to remove scroll when the menu is open 

3. Also, the menu should stretch to the full width and height of the browser window.

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2025-01-31 at 13 12 24
- Check styles here

VitaliyBondarenko1982 (approved):
Good job.

PR #1041, opened 2025-01-29
add tast solution

Slavko3111:
[DEMO LINK](https://Slavko3111.github.io/layout_dia/)


etojeDenys (approved):
it's better to disable the scroll if menu opened
image

PR #1043, opened 2025-01-30
Develop

VladTk:
[DEMO LINK](https://VladTk.github.io/layout_dia/)


etojeDenys (approved):
good job, but it's better to disable scroll if menu opened

PR #1047, opened 2025-02-04
Develop

VSkits:
[DEMO LINK](https://VSkits.github.io/layout_dia/)


VSkits:
need some help with max-width of slider block
in task on vp-width > 1600px must not spread
IDK how!)))
and after realise slider effect

volodymyr-soltys97 (changes requested):
Hi, if you need some help with slider pls feel free in the fe_chat 

Also, add cursor pointer for all clickable elements 
image

Width for these blocks should be responsive 
image

Remove scroll when the menu is open 
image

VSkits:
i have some troubles with transition in slider block
i found som information about transition of display from none to block (it doesnt work)
so i have no solutions in my head

anastasiiavorobiova (changes requested):
Great work!
- The layout should match the design on breakpoints
- Screenshot 2025-02-05 at 14 48 26
- Screenshot 2025-02-05 at 14 46 58
- There is no need to open the new tab after the click on the phone

VSkits:
@anastasiiavorobiova 
but how?
tel. number should work
and about width
in task:
"on large screens limit the width of black container to 1600px, so that left outer space have a standart grey background color"

volodymyr-soltys97 (approved):
Good job 👍

PR #1048, opened 2025-02-05
dia-air

olga-wukelan:
[DEMO LINK] https://olga-wukelan.github.io/layout_dia/


etojeDenys (changes requested):
images looks different
image

2. center the button
image


3. header's height need to be only 100vh
image

PR #1054, opened 2025-02-17
add solution

KOliinyk:
[DEMO LINK](https://KOliinyk.github.io/layout_dia/)


maxim2310 (changes requested):

1. need to fix slider and horizontal scroll 
image
image

2. add smooth scroll after click on anchor link 
3. add cursor pointer and transition effect on all interactive elements such as 
image
4. remove autofill style 
image

PR #1056, opened 2025-02-19
add solution

KOliinyk:
[DEMO LINK](https://KOliinyk.github.io/layout_dia/).


maxim2310 (changes requested):

1. first block should have 100vw width 
image
2. another blocks should be centered 
image
3. add smooth scroll while clicking on anchor links 
4. add transition on all animations especially on buttons
image
5. here you have transition only on enter but not on out 
image
6. same on mobile 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The image should fit to the right side of the browser window.
image

2. Not fixed from the previous review 
image

PR #1060, opened 2025-02-25
solution

RusikS98:
[DEMO LINK](https://RusikS98.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Change the autofill styles, the autofill and default styles should be are same
image

2. Make all these fields are required before sending 
image

3. Fix the font family, check it on the design 
image

4. Add the topical address for these social networks 
image

vadiimvooo (changes requested):
Looks like comment with autofill is not fixed. Please feel free to ask for help in fe_chat

RusikS98:
> Looks like comment with autofill is not fixed. Please feel free to ask for help in fe_chat

> I think I changed, now it shows autofill text diff color, has border and background; It should be correct.

etojeDenys (approved):
disable the scroll if menu is open
image

PR #1061, opened 2025-03-04
Dia

bbbozhena:
https://bbbozhena.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

image

3. The image in slider is too big, need to use width 50% here 
image

4. Change the autofill styles for the form fields. The autofill and default styles should be are same
image

5. This element should be link 
image

6. Add hover effects here 
image

7. Need to implement the burger menu when user clicking here. Design of your choice
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The horizontal indents should be are same everywhere 
image
image
image

2. After submit the form don't need scroll page to top. You need only reset form fields 
image

3. Not fixed from the previous review 
image

maxim2310 (changes requested):

1. buttons are not round 
image
2. something wrong with arrow 
image
3. photos are stretched out
image
4. indents should be the same 
image

VitaliyBondarenko1982 (approved):
Looks good to me

PR #1065, opened 2025-03-17
add task solution

VictorKomara:
https://VictorKomara.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements on the page. Also for cards
image
image

2. These elements should be links 
image
image

3. Make these fields are required before submit 
image

vadiimvooo (approved):
Well done.

PR #1069, opened 2025-03-25
add task solution

DmutroSheva:
[DEMO LINK](https://DmutroSheva.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check after the demo 
image

PR #1074, opened 2025-04-06
Develop

viktoriamyhailiak:
Demo: https://viktoriamyhailiak.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to fix width for slider images, check it on the design 
image

2. After submit the form don't need scroll page to top. 
image

3. Add transition when this menu appears on the page 
image

viktoriamyhailiak:
As I understand from the design, I need to make a slider with full width on tablet and mobile version. Then 502px width on small-desktop(1024) and 687px on large-desktop(1600). I made the width on large desktops on 1600(not 1280px), as the content of the main page doesn't fit in the free space and starts to overflow because of the padding-inline. And the final width of the slider on large screens is 1167px. Do I understand this correctly?

volodymyr-soltys97 (approved):
Great! 🔥

PR #1075, opened 2025-04-07
created air landing

MykolaFatkullin:
[DEMO LINK](https://MykolaFatkullin.github.io/layout_dia/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The image in the slider is very wide, you need to use a width of 50% of the browser window width
image

image

3. Need to open these additional links in the new tab
image

4. The close icon should be white and add some gaps for nav links 
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #1077, opened 2025-04-17
Dia

AndrewT98:
[DEMO LINK](https://AndrewT98.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements 
image

2. The user cannot be able to change size of the textarea field 
image

3. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

4. Add hover effect here 
image

5. These elements should be links 
image

anastasiiavorobiova (approved):
Great work!

PR #1087, opened 2025-05-09
add task solution

Andrey1166:
[DEMO LINK](https://Andrey1166.github.io/layout_dia/).


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. An external links should be open in the new tab 
image

2. Don't use max width for the content, use `width: 100%` instead 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
Make header by design on full screen width as footer. 
Screenshot 2025-05-10 at 09 05 43

vadiimvooo (approved):
Well done

PR #1095, opened 2025-06-03
Develop

PAVLO111:
-DEMO LINK- https://PAVLO111.github.io/layout_dia/


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add hover effects for all clickable elements on the page 
image

2. Change the autofill styles for the form fields. The autofill and default styles should be are same 
image

3. These elements should be links and add hover effects here 
image

4. Add scroll behavior smooth to the page

5. Need to implement the burger menu when user clicking here 
image

etojeDenys (changes requested):
fix the styles, then request a review again
image

etojeDenys (changes requested):
fix the demo link, and then request a review again
image

etojeDenys (changes requested):
Good job
To improve:
 1. all input should be required
image

2. the button should reset the inputs instead of the page

3. add some hover to buttons

4. the menu does not show up when the button is pressed
image

etojeDenys (approved):
lgtm

PR #1105, opened 2025-06-28
add solution

IvanRyabukha:
https://IvanRyabukha.github.io/layout_dia/


etojeDenys (changes requested):
1. after clicking header logo we see error
image
image

2. after entering the data, we still see an error
image

3. it's better to show the error on each of the fields after pressing the button

etojeDenys (approved):
lgtm

PR #1117, opened 2025-09-07
add solution

SashaGlazko:
[DEMO LINK](https://SashaGlazko.github.io/layout_dia/).


Anton-Kuchmasov (changes requested):
Great job you did!

However, there are a couple of issues need to be fixed:
1. Please change the [autofill styles](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill)
image

2. Could you reduce `scale` a little bit?
image

etojeDenys (approved):
good job

PR #1128, opened 2025-11-08
mySolution

AntonFof:
[DEMO LINK](https://AntonFof.github.io/layout_dia/)


h-amster (changes requested):
1. Add cursor: pointer to interactive buttons
image
image

2. Right now, when you hover over elements, multiple properties change, but the transition only applies to one of them.
Please fix it so that the transition also affects the box-shadow when hovering over the cards — currently, it appears instantly. Also, make the scale animation for the social media icons smooth as well.
image
image
image

3. Reduce the scale value for the buttons a bit, since on narrow screens the hover effect causes a horizontal scrollbar to appear.

image

PR #1138, opened 2025-12-22
solution

kazonni:
[DEMO LINK](https://kazonni.github.io/layout_dia/)


2pasha (changes requested):
good job!

here are some improvements:

1. when user submit the form - page shouldn't be scrolled up

2. add hover effects for these elements
image

3. add transition for these elements hover effect
image

Anton-Kuchmasov (changes requested):
When User selects one of previously added input values (e.g. mail address / name), all styles of autofill should be the same as usual.
image

Check out [this link](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) to get more info and fix it in your solution

Also pay attention for `Send` button behavior. No scroll-ups or page reloads - just wipe all entered data and it'll be enough.
Check out #15 of [this](https://github.com/ArtemNosachenko/layout_landing-page/blob/develop/checklist.md) instruction

vitalii-bondarenko-ua (approved):
Looks good

PR #1140, opened 2026-01-04
add task solution

dimon2202:
[DEMO LINK](https://dimon2202.github.io/layout_dia/)


FaiHamid (changes requested):
Good job!
Let`s just fix few details:

1.On the desktop, content must have width restrictions.
image

2.The menu doesn't open
image

etojeDenys (approved):
good job. just one thing to fix

1. it should be tel link

Image

PR #1141, opened 2026-01-08
Develop

annaskobara:
[DEMO LINK](https://annaskobara.github.io/layout_dia/).


2pasha (changes requested):
please make sure that you have deployed your app 

because right now i can see error 404
image


if you need some help - feel free to ask your questions in fe_chat 💬

Denys-Kravchuk9988 (changes requested):
it looks like the styles didn’t load correctly. 
image
image

In this case it can be caused by "homepage" property
image

It's needed to set this value as link for deployed project (this property is in `package.json` file)
`"homepage": "https://annaskobara.github.io/layout_dia"`

after this change you need to re-deploy the project

If it doesn't help, please feel free to ask in chat

etojeDenys (approved):
good job

PR #1145, opened 2026-01-24
add task solution v1

anari-a:
https://anari-a.github.io/layout_dia/

## Checklist for preparing a portfolio project for HR review

1. Don’t forget to add the title “Air” for the whole web page
2. A landing page is implemented strictly according to the design in Figma
4. Links in the header and footer menus should lead to the corresponding blocks of the landing page
5. The speed of animations is the same throughout the landing page (for example, increasing when hovering or moving blocks when scrolling)
6. Placeholders in the forms suggest what to enter, and if there is a validation of the form, then it is clear in what format to enter the phone number
7. Make sure everything looks neat on mobile and without horizontal scrolling
8. Add favicon
9. Add a smooth scroll for the whole page
10. When you try to send the form there is no 405 error and the form is automatically cleared after submit and is scrolled to the top of the page or the page is reloaded
11. The form shouldn’t submit empty
12. The "Learn more" button should lead to the closest block (Our expertise)
13. The "Hire Us" button in the header menu should lead to the contact form
14. In the PROCESS section add a hover effect to cards (for example, let them increase in size a little)
15. Inside the cards in PROCESS sections, make a link out of the "Learn More" text, so that it hovers simultaneously with the arrow button on the left
16. In the Vision, Passion, Results section, the “Apply” button also should lead to the contact form
17. In the Testimonials and Our expertise section add hover effects for each review
18. After clicking the menu buttons, the landing page will scroll exactly to the beginning of the desired block (exactly where its section begins)
19. The arrows at the bottom of the first block should not respond slowly to pressing (this is possible if the pictures are too large)
20. When a user clicks on Address Google Maps is opened in a new tab
21. The user must have the opportunity to conveniently write 2-3 lines of text in the message field
22. OPTIONAL: After everything is done, you can implement the slider on the hero page (you can find images for it on Unsplash)


etojeDenys (changes requested):
1. it would be nice to change the colour of the links when you hover over them
image

2. add a cursor pointer to the buttons

3. page looks broken right now
image

anari-a:
Hi, thanks for such a quick response. I have a few questions about it, could you please help me with it?

1. I’ve corrected the code, so now the links  should change their color to blue, but this only applies to the phone-number and address links. The issue with the social media icons is that we don't have them in blue in Firma, so I’m not sure how to change their color, since they are images I downloaded from Figma. So I just make them a little bit bigger when hover.

2. I’ve changed the cursor to a pointer, no problems here, I hope.

3. I do notice that the page seems broken when I open it on my Mac. However, the task was to style the page for fixed screen sizes: 320, 640, 1024, 1280, 1600, and 2560px. When I check these screen sizes via Chrome’s developer tools, everything looks neat and fine. Could you please explain briefly what I need to change here if any adjustments are still necessary?

etojeDenys (changes requested):
the page should look good at every width. the breakpoints - 320, 640, 1024, 1280, 1600 and 2560 px - help you understand how the page should change and adapt to different screens

if you have any questions, it's better to ask in the chat because it's difficult to communicate here

PR #1147, opened 2026-01-27
Layout_dia Donegit add src/styles/utils/_mixins.scss

tiagodearaujo13-lab:


IanLuan (changes requested):
Hi Tiago! There are some style issues. It looks like the footer styles are duplicated in forms.scss, and there are also some errors in the class names, so the styles are not being applied correctly (https://tiagodearaujo13-lab.github.io/layout_dia/). Please review the styles and run the deploy again. If everything is OK, click on “Request review” again.

IanLuan (approved):
Good job, Tiago! I’m approving your task, but I’ll leave a few tips to help you improve the project:

- Browsers autocomplete fields based on the `name` attribute. If your email input has `type="email"` but doesn’t have `name="email"`, the browser won’t suggest saved emails. So always make sure your inputs have the correct name.
- The container of the “Our Expertise” section is wider than the other sections. It should have the same width. I suggest using `max-width` to control the container sizes consistently.
- The logo image in the footer is not working.

Nice job again!

PR #1151, opened 2026-02-04
Develop

kristinakovalchuk:
[DEMO LINK](https://kristinakovalchuk.github.io/layout_dia/).


etojeDenys (changes requested):
great job

1. this link should open in a new tab
Image

2. menu does not open when the button is pressed
Image

[here](https://cumacea.github.io/layout_dia/#home) is an example of how you can implement it

etojeDenys (approved):
good job

PR #1159, opened 2026-02-26
subindo trabalho

JohannPDaniel:


src/index.html
```diff
@@ -0,0 +1,458 @@
...
+    <title>Air Black</title>
+    <link
+      rel="stylesheet"
+      href="./styles/main.scss"
+    />
+  </head>
+  <body>
+    <header class="header__nav">
+      <div class="header__nav--logo"></div>
+      <ul class="header__nav--navbar">
```
IanLuan, line 18: Use a nav element for the navigation list.

IanLuan (changes requested):
Great progress! You are almost there! There are still a few things that need to be adjusted according to the checklist:

1- Use a nav element for navigation links
2- Links in the header and footer menus should lead to the corresponding blocks of the landing page
3- Add favicon
4- The form shouldn’t submit empty
5- The "Learn more" button should lead to the closest block (Our expertise)
6- The "Hire Us" button in the header menu should lead to the contact form
7- In the PROCESS section add a hover effect to cards (for example, let them increase in size a little)
8- Inside the cards in PROCESS sections, make a link out of the "Learn More" text, so that it hovers simultaneously with the arrow button on the left
9- In the Vision, Passion, Results section, the “Apply” button also should lead to the contact form
10- In the Testimonials and Our expertise section add hover effects for each review
11- After clicking the menu buttons, the landing page will scroll exactly to the beginning of the desired block (exactly where its section begins)
12- When a user clicks on Address Google Maps is opened in a new tab

PR #1161, opened 2026-03-17
finish layout dia

sonik-boom71:
- https://sonik-boom71.github.io/layout_dia/


etojeDenys (changes requested):
links do not work. please fix them and request the review again

PR #1163, opened 2026-03-24
feat: Landing Page - Layout dia

wiliammunchen:


IanLuan (changes requested):
Great job! The page is looking very close to the mockup. I have only one small request: In the Hero section, the slider is currently below the text, but on desktop, it should be on the right. It seems like the mobile adjustment might have affected the desktop layout as well.

IanLuan (approved):

PR #1172, opened 2026-04-19
add task solution

vitamartsekhovska:
[DEMO LINK](https://vitamartsekhovska.github.io/layout_dia/).


vitalii-bondarenko-ua (changes requested):
Good job.
- All form fields should be required and have correct types (email, tel, etc.)
- Placeholders should give examples of what should be added, to make the expected format clear
- The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
`<form onsubmit="this.reset(); return false;">`. Actually now nothign happens on click to send button - for fix it button should be inside form tag.

Image

etojeDenys (approved):
good job

PR #1174, opened 2026-04-24
solution

AngelaStrechen:
[DEMO LINK](https://AngelaStrechen.github.io/layout_dia/)


Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. I would recommend to remove autofill
Image

2. It's better to disable scrolling when menu is open
Image
Image

3. On tablet the page looks broken.
Image

2pasha (changes requested):
comments from previous review is still actual, make sure that you have redeployed your app

also some more things to improve: 

1. add favicon 

Image

2. i would recommend to remove `box-shadow` value on hover

Image

2pasha (changes requested):
good progress ! 👏

here are last improvements: 

1. disable page scrolling when menu is opened. try this approach 

Image

2. when user click on nav items - page should be scrolled smoothly 

Image

etojeDenys (approved):
good job

PR #1176, opened 2026-04-29
add solution

RuslanV23:
https://ruslanv23.github.io/layout_dia/


etojeDenys (changes requested):
almost done

1. send button should clear the fields without submitting

2. disable page scrolling when the menu is open
Image

etojeDenys (approved):
good job

PR #1178, opened 2026-05-05
solution

NeustrVk:
[DEMO LINK](https://NeustrVk.github.io/layout_dia/)


brespect (changes requested):
Good progress, but seem like your DEMO LINK is not working:
Image

PR #1180, opened 2026-05-09
add solution

Evg3n22:
[DEMO LINK](https://Evg3n22.github.io/layout_dia/)


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. I would recommend to implement some menu (you can take from other landing design)
Image

2. It's better to add links for these socials (not personal, just for main page https://www.instagram.com/ for example)
Image 

3. I would recommend to add `cursor: pointer` on these places
Image
Image

4. Here is better to remove autofill styles
Image

5. Learn more should have Poppins as font
Image
Image

6. There are needed spaces according to design
Image
Image

brespect (changes requested):
Good progress, check next:
1. Still not fixed: when I select the email autofill is still styled.

2. I can't submit the form:Image

3. Slider buttons seem not working.

4. Add hover to all interactive elements: buttons, links, pictures etc.

etojeDenys (changes requested):
1. this link should open in a new tab

Image

2. every field should be required

Image

3. this input should be of the 'email' type

Image

4. font looks different here

Image

5. there is too much space here

Image

2pasha (changes requested):
good job! 👏

here are last improvements: 

1. add favicon 

Image

2. i suppose here should be another font 

Image

3. after form submitting - page should not be refreshed, scrolled, etc. Take a look on instructions [here](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

4. it will be great if you implement menu opening functionality 

Image

brespect (approved):
Looks good 😊

PR #1184, opened 2026-05-16
first-check

notbelievable:
[DEMO LINK](https://notbelievable.github.io/layout_dia/)


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. I would recommend to add `cursor: pointer` for all block because by hover outside the image there is also scaling
Image

2. This animation seems to be broken. By hover on this section there are lags
Image

3. It's better to add `cursor: pointer` by hover this text
Image

4. It's needed to remove autofill styles
Image

Anton-Kuchmasov (changes requested):
Good work, but there are a few issues need to be fixed:

1. Fix typo mistake: there should be `leaRn more`:

Image

2. Also, please slightly reduce `hover` effect to avoid cards overlapping

3. On mobile view, aside menu is not working :(

Image

brespect (approved):
Looks good

PR #1191, opened 2026-06-04
Develop

RSydorenko322:
https://rsydorenko322.github.io/layout_dia/


etojeDenys (changes requested):
almost done, just two small fixes:
1. there is too much space
Image

2. disable text area resizing

etojeDenys (approved):

PR #1202, opened 2026-08-05
Layout-dia

olegpuzin:
[DEMO LINK](https://olegpuzin.github.io/layout_dia/).


2pasha (changes requested):
good job! 👏

here is one improvement: 

after form submit - form just should be reseted, without page redirects, scrolls etc

brespect (changes requested):
Make sure, you have deployed your solution, got this:
Image

roman-mirzoian (changes requested):
The deployment process did not work properly, you need to figure out what the problem is:
Image

Anton-Kuchmasov (changes requested):
Not fixed yet :(

Image

Steps to reproduce:
1. fill `Form` with some valid data (name, email, message)
2. click `Send` button
3. observe 405 Error


however, shall it be some troubles with deploy?
currently now (Aug, 6th, 2026, 18:49 UTC +3 Kyiv Summer Time), there's some degration with GH Actions & GH Pages deployment

Image

you can check the status on [GitHub Status page](https://www.githubstatus.com/) and when it'll be okay - re-deploy your solution

etojeDenys (changes requested):
almost done:
1. it would be better to centre the content of this section on smaller screens


Image

2. make sure that this section looks the same as the layout
example:

Image

your(too much space here):

Image

brespect (approved):
Looks good!

